PR #659, opened 2014-12-29
[jsk_tilt_laser] Add use_robot_description argument to multsense.launch

orikuma:
- Add use_robot_description arg to multisense.launch to prevent robot_description from conflicting with one of real robot
- Removed robot_description private param in ros_driver, which is seemed to be unused in multisense_ros


garaemon:
ticket written
https://bitbucket.org/crl/multisense_ros/issue/32/add-argument-to-disable-to-set

PR #688, opened 2015-01-12
[image_view2] Do not use time difference to detect point or rectangle

garaemon:


garaemon:
we don't need to wait for 0.5~1.0 sec to select rectangular region.

PR #690, opened 2015-01-13
use jsk-ros-pkg/archives for download tgz

k-okada:
this will revert #676


garaemon:
using github as archive, looks good

PR #692, opened 2015-01-13
[jsk_tilt_laser] Add scan_to_cloud_chain to multisense.launch to get one-scan pointcloud.

garaemon:
We use ~high_fidelity=true in order to avoid
laser_geometry's bug to produce large sphere pointcloud

* sphere bug when we use ~high_fidelity=false

sphere-bug


garaemon:
~high_fidelity=true

nonsphere

PR #723, opened 2015-01-30
Remove rosbuild files

garaemon:
* remove manifest.xml

```
find . -name manifest.xml | xargs -n 1 git rm
```

* rename catkin.cmake

```
find . -name catkin.cmake -exec bash -c git rm `dirname {}`/CMakeLists.txt; git mv {} `dirname {}`/CMakeLists.txt \;
```


k-okada:
some of Makefike is no longer required

garaemon:
I will check it

garaemon:
Removed All the makefiles for rosbuild

PR #725, opened 2015-01-31
[jsk_tilt_laser] Update multisense.launch according to the latest update

garaemon:
@YoheiKakiuchi @orikuma @terasawa Be careful about this update

1) use multisense.launch, it support launch_robot_state_publisher argument
2) fix name to change speed of spindle laser


garaemon:
https://bitbucket.org/crl/multisense_ros/issue/32/add-argument-to-disable-to-set

PR #733, opened 2015-02-03
Do not test source compilation

garaemon:


garaemon:
only deb package is enough and we don't want to run so many tests because it delays other tests...

PR #736, opened 2015-02-04
NLopt CMakeLists.txt

cbames:
When I attempted to use the nlopt hydro package, the nlopt library was not being linked by packages that depended on it. I believe this is because catkin_package did not indicate that the library was being exported.

Thus I added nlopt_cxx library to catkin_package so that the library will be included by packages that depend on nlopt.


k-okada:
Great, thanks

PR #739, opened 2015-02-05
Add rosbag2csv

furushchev:
add rosbag csv exporter since rqt_bag no longer supports csv export.

### DIFF
--- a/jsk_topic_tools/scripts/bag2csv.py
+++ b/jsk_topic_tools/scripts/bag2csv.py
@@ -94,7 +94,7 @@ def bag_to_csv(options, args):
         bag.close()
 
 if __name__ == '__main__':
-    rospy.init_node('bag2csv')
+    rospy.init_node('bag2csv', anonymous=True)
     parser = OptionParser(usage="%prog [options] bagfile")
     parser.add_option("-a", "--all", dest="all_topics",
                       action="store_true",
--- a/jsk_topic_tools/scripts/bag2csv.py
+++ b/jsk_topic_tools/scripts/bag2csv.py
@@ -104,4 +104,8 @@ if __name__ == '__main__':
                       help="end time of bagfile", type="float")
     (options, args) = parser.parse_args()
 
+    if len(args) != 1:
+        parser.print_help()
+        exit(0)
+
     bag_to_csv(options, args)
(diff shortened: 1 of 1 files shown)


jsk_topic_tools/scripts/bag2csv.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    stream: StringIO
+    msg: message
+    """
+    try:
+        for s in type(msg).__slots__:
+            val = msg.__getattribute__(s)
+            message_type_to_csv(stream, val, ".".join([parent_content_name,s]))
+    except:
+        stream.write(", " + parent_content_name)
```
garaemon: I think no space is required after `,`

garaemon:
`rostopic echo -p` is enough?

```
$ rostopic echo hoge -p
%time,field.data
1423155837204827070,hoge
1423155838204751014,hoge
1423155839204807043,hoge
```

garaemon:
rostopic also can read bag file with `-b` option.
```
Usage: rostopic echo [options] /topic

Options:
  -h, --help            show this help message and exit
  -b BAGFILE, --bag=BAGFILE
                        echo messages from .bag file
  -p                    echo in a plotting friendly format
  --filter=FILTER-EXPRESSION
                        Python expression to filter messages that are printed.
                        Expression can use Python builtins as well as m (the
                        message) and topic (the topic name).
  --nostr               exclude string fields
  --noarr               exclude arrays
  -c, --clear           clear screen before printing next message
  -a, --all             display all message in bag, only valid with -b option
  -n COUNT              number of messages to echo
  --offset              display time as offsets from current time (in seconds)

```

k-okada:
Please contribute to rqt_bag, we need to escape from infinite loop ->
http://answers.ros.org/question/9102/how-to-extract-data-from-bag/?answer=63877#post-id-63877

garaemon:
I don't think rqt_bag should have this feature. rostopic is enough

✉︎ Ryohei Ueda




On Thu, Feb 5, 2015 at 5:10 PM -0800, "Kei Okada" <notifications@github.com> wrote:












Please contribute to rqt_bag, we need to escape from infinite loop ->

http://answers.ros.org/question/9102/how-to-extract-data-from-bag/?answer=63877#post-id-63877

k-okada:
if rostopic can handle this , that it much much better,

--
◉ Kei Okada

PR #743, opened 2015-02-06
[jsk.rosbuild] Use ros wrapper of hrpsys and openrtm and checkout euslisp, jskeus and geneus before run "rosdep install"

garaemon:
1. Use ros-wrapper of hrpsys and openrtm instead of non-catkin hrpsys and openrtm. 

  Because the packages are not yet ready to use it.
  Need to solve [this issue](https://github.com/start-jsk/rtmros_common/issues/651)

2. Run rosdep install after call `jsk_roseus/setup.sh` not to install euslisp, jskeus and geneus.

### DIFF
--- a/jsk.rosbuild
+++ b/jsk.rosbuild
@@ -47,7 +47,7 @@ while [ -n "$1" ] ; do
 	*) echo "Unknown option($1)"; usage;;
     esac
 done
-
+echo USE_SOURCE="$USE_SOURCE"
 # set environment variables
 # set distribution
 case $1 in
@@ -131,7 +131,7 @@ function setup-ros {
     if [[ $REPLY =~ ^[Yy]$ ]]; then
         $COMMAND sudo apt-get update
     fi
-    $COMMAND sudo apt-get upgrade $YES 
+    $COMMAND sudo apt-get upgrade $YES
     $COMMAND sudo apt-get -qq $YES install ros-${DISTRIBUTION}-rosbash python-rosdep python-wstool python-catkin-tools python-pip
     if [ "$USE_SOURCE" == "full" ]; then
         $COMMAND sudo apt-get -qq $YES install python-rosinstall-generator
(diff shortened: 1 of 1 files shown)


k-okada:
https://github.com/k-okada/jsk_common/blob/use_catkin_tools/jsk.rosbuild would be dirty workaround for this issue.

fundamental problem is to decide whether  if hrpsys is catkin package ? or non-catkin package.
If we decide to consider hrpsys as non-catkin package, then we nolonger able to use `rosrun hrpsys ProjectGenerator`,  so you can not use them within launch file. 

At this moment if you install hrspys as deb, it installed under /opt/ros/indigo/share, so you can use `rosrun hrpsys ProjectGenerator` but it is just a side effect that /opt/ros/indigo/setup.sh set ROS_PACKAGE_PATH to /opt/ros/indigo/share, so we may have to create ProjectGenerator under hrpsys_tools package and use them

garaemon:
> fundamental problem is to decide whether if hrpsys is catkin package ? or non-catkin package.

Yes, this is the problem.

> At this moment if you install hrspys as deb, it installed under /opt/ros/indigo/share, so you can use rosrun hrpsys ProjectGenerator but it is just a side effect that /opt/ros/indigo/setup.sh set ROS_PACKAGE_PATH to /opt/ros/indigo/share, so we may have to create ProjectGenerator under hrpsys_tools package and use them

I'd like to use this script to set up computers inside of robots like hrp2 v. In that case, I strongly want to
install hrpsys and openrtm from source code.


I cannot agree with using non-catkin hrpsys in any way until [this issue](https://github.com/start-jsk/rtmros_common/issues/651) is solved. 
I cannot understand why we need to have script to install package which does not work.

k-okada:
or use http://answers.ros.org/question/54640/roslaunch-external-program/ to start openrtm/openhrp3/hrpsys program, but it seems gazebo teams choose to create gazebo_ros package and start gazebo program via that package.

garaemon:
> or use http://answers.ros.org/question/54640/roslaunch-external-program/ to start openrtm/openhrp3/hrpsys program, but it seems gazebo teams choose to create gazebo_ros package and start gazebo program via that package.

It looks we cannot resolve `$(find hrpsys)` in launch files by this technique.

k-okada:
yes, so we have to change every `find hrpsys` to find `hrpsys_tools` and update  hrpsys_tools to work.

PR #759, opened 2015-02-10
[jsk_tools] Install bin directory to lib directory

garaemon:


garaemon:
sourceforge is down now...

PR #761, opened 2015-02-10
[jsk_topic_tools] Add ~latch option to snapshot nodelet

garaemon:


garaemon:
recently indigo testing never succeeds

Issue #772, opened 2015-02-12
setup.py breaks build


dirk-thomas:
http://54.183.26.131:8080/job/Idoc__jsk_common__ubuntu_trusty_amd64/2/consoleFull

I guess the problem is in the setup.py: https://github.com/jsk-ros-pkg/jsk_common/blob/c7eff2008e8d72ce4e3a6b625a15557c4b44efaa/bayesian_belief_networks/setup.py#L8-L11

The code is:
* very fragile: `sys.version[:3]` will break for version numbers like *3.10*
* the decision about dist-packages / site-packages can not be made here since catkin invokes the setup.py file in a different way and make the decision

garaemon:
Thanks!

@aginika Please take care of this issue

PR #780, opened 2015-02-15
[bayesian_belief_networks] Fix setup.py for installing python

aginika:
From discussion #772 
Still not solve the problem of dist-packages / site-packages.
This just solve the problem of version number.


aginika:
According to https://github.com/jsk-ros-pkg/jsk_common/issues/772#issue-57502125,
```
the decision about dist-packages / site-packages can not be made here
 since catkin invokes the setup.py file in a different way and make the decision
```
So we can't decide dist or site in setup.py?

PR #824, opened 2015-03-15
set ROS_DISTRO explicitly

eisoku9618:


k-okada:
Why do we have to set ROS_DISTRO

eisoku9618:
because ros_distro may be set to old version , for example when we upgrade
from groovy to hydro.

https://github.com/jsk-ros-pkg/jsk_roseus/blob/master/setup_upstream.sh#L62
2015/03/16 9:53 "Kei Okada" <notifications@github.com>:

k-okada:
i see

Issue #834, opened 2015-03-29
Couldn't find oracle java7 which is written in jsk.rosbuild


aginika:
https://github.com/jsk-ros-pkg/jsk_common/blob/master/jsk.rosbuild#L118

I couldn't find oracle-java7.

garaemon:
Please remove that, we no longer use android stuff

aginika:
Thanks a lot.

PR #835, opened 2015-03-29
Remove oracke-java7 from jsk.rosbuild

aginika:
From #834


k-okada:
please update to use this ppa and install oracle-java7-installer

http://www.webupd8.org/2012/01/install-oracle-java-jdk-7-in-ubuntu-via.html

aginika:
I updated

k-okada:
i think this requires user's input, please confirm.

PR #851, opened 2015-04-08
[jsk.rosinstall] Add ar_tools repository

garaemon:


k-okada:
FYI: I heard ar-track-alvar is better than this, if you have troubl e
on ar-tools, please try.
https://github.com/tork-a/live-cd/blob/master/00-create-cd.sh#L111
--
◉ Kei Okada

PR #853, opened 2015-04-09
[image_view2] add grid option

aginika:
With param, this will show the grid.


garaemon:
Please add dynamic_reconfigure API to change parameter

PR #868, opened 2015-04-20
[jsk_tools] Add rossetdefault, rosdefault to zshrc.ros

wkentaro:
Usage
-----
```
user@host $ rossetdefault pr1040
set ROS_MASTER_URI to http://pr1040:11311
[http://pr1040:11311] user@host $
user@host $ rosdefault
set ROS_MASTER_URI to http://pr1040:11311
[http://pr1040:11311] user@host $
user@host $ rossetdefault local
user@host $ rosdefault
set ROS_MASTER_URI to http://localhost:11311
```
It is recommended to write `rosdefault` in your .bashrc or .zshrc.

### DIFF
--- a/jsk_tools/env-hooks/99.jsk_tools.bash
+++ b/jsk_tools/env-hooks/99.jsk_tools.bash
@@ -1,6 +1,27 @@
 #!/bin/bash
 # -*- mode: Shell-script; -*-
 
+function rossetdefault() {
+    local hostname=${1-"local"}
+    local ros_port=${2-"11311"}
+    echo "$hostname\n$ros_port" > ~/.rosdefault
+    rosdefault
+}
+function rosdefault() {
+    if [ -f ~/.rosdefault ]; then
+        local hostname="$(sed -n 1p ~/.rosdefault)"
+        local ros_port="$(sed -n 2p ~/.rosdefault)"
+    else
+        local hostname="local"
+        local ros_port="11311"
+    fi
+    if [ "$hostname" = "local" ]; then
+        rossetlocal $ros_port
+    else
+        rossetmaster $hostname $ros_port
+    fi
+}
+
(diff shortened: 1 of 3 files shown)


k-okada:
please take care of both bash and zsh settings, using different command
between zsh and bash users may confusing.

--
◉ Kei Okada

wkentaro:
okay I will also add this function to bashrc.ros.

garaemon:
I think there is a lot of similar function between zsh and bash, Can we use same file?
for example add `source $(rospack find jsk_tools)/common.sh`

wkentaro:
I think it is possible and not so hard.
I opened issue #869

jsk_tools/env-hooks/99.jsk_tools.bash
```diff
@@ -1,6 +1,21 @@
...
 # -*- mode: Shell-script; -*-
 
+function rossetdefault() {
+    local ros_master_default=${1-"local"}
+    echo $ros_master_default > ~/.rosdefault
+    rosdefault
+}
+function rosdefault()
+{
+    local ros_master_default=`cat ~/.rosdefault`
```
garaemon: Check if ~/.rosdefault exists

wkentaro:
I updated.

garaemon:
I'm afrait `rossetdefault` may be too complicated utility but I merge it in order to check if it's useful.

wkentaro:
I see.

It may become simple command by thinking inversely, 

`rosdefault` is the command set `ROS_MASTER_URI` with values written in `~/.rosdefault`.

And

`rossetdefault` is the command by which you can easily rewrite `~/.rosdefault`.

k-okada:
it think this kind of files should go to ~/.ros ($ROS_HOME)http://wiki.ros.org/ROS/EnvironmentVariables#ROS_HOME

PR #876, opened 2015-04-24
[downward] use xzf to extract .tar.gz, instaed of xvzf to reduce log len...

k-okada:
...gth


garaemon:
BTW, we should use [externalproject_add](http://www.cmake.org/cmake/help/v3.0/module/ExternalProject.html) for this kind of task

garaemon:
ar_tools use [externalproject_add](https://github.com/ar-tools/ar_tools/blob/master/artoolkit/CMakeLists.txt#L16)

k-okada:
thanks, I did not know that.

--
◉ Kei Okada

PR #879, opened 2015-04-25
[jsk_tools] save rosdefault file under ROS_HOME

wkentaro:


k-okada:
I had trouble on this issue, because the robot's bashrc is written as 
> rossetdefault

and someone uses this terminal with different robot name for temporally, and after that I opened my terminal and the ROS_MASTER_URI was changed.

How about add, `--save` option, if we need to store "default value", we need to explicitly tell so.

wkentaro:
It is a wrong usage.
rosdefault is written in bashrc.
and

- rossetdefault pr1012 for saving config (as u say with --save option)
- rossetmaster pr1012 for temporary use

k-okada:
I see... 
@h-kamada 
http://jsk-docs.readthedocs.io/en/latest/jsk_common/doc/jsk_tools/cltools/setup_env_for_ros.html?highlight=rossetmaster

PR #881, opened 2015-04-26
[jsk_tools] now you can install pygithub3 by rosdep install

wkentaro:
```
rosdep update
roscd && cd ..
rosdep install --from-paths . -y -r
```


garaemon:
great!

Issue #885, opened 2015-04-26
error when source setup.zsh


wkentaro:
```
% source ~/ros/hydro/devel/setup.zsh
/home/wkentaro/ros/hydro/devel/etc/catkin/profile.d/99.jsk_tools.zsh:source:6: no such file or directory: /home/wkentaro/ros/hydro/devel/99.jsk_tools.sh
```

In indigo environment it does work, but in hydro I have this error.

wkentaro:
I found how to solve this problem.
You should manually edit setup.zsh you `source`.
It should be like below::
issue in catkin repo: https://github.com/ros/catkin/issues/686
```sh
#!/usr/bin/env zsh
# generated from catkin/cmake/templates/setup.zsh.in

CATKIN_SHELL=zsh
_CATKIN_SETUP_DIR=$(builtin cd -q "`dirname "$0"`" > /dev/null && pwd)
# emulate sh # emulate POSIX
# . "$_CATKIN_SETUP_DIR/setup.sh"
# emulate zsh # back to zsh mode
emulate -R zsh -c 'source "$_CATKIN_SETUP_DIR/setup.sh"'
```
And this will solve the some problems in your zsh configuration like::
```sh
% ..
zsh: permission denied: ..
```

PR #888, opened 2015-04-27
[jsk_tools] temporary change to avoid error caused by bug in ros/catkin repo

wkentaro:


wkentaro:
issue: https://github.com/jsk-ros-pkg/jsk_common/issues/885

PR #904, opened 2015-05-01
[jsk_data] Add KEYWORD to large-list/small-list target in Makefile

wkentaro:
**Usage**
```
USER=wada make large-list KEYWORD=jsk_2014_picking_challenge
```


k-okada:
nice. please dd that Usage into the makefile

wkentaro:
I added.

PR #909, opened 2015-05-04
[jsk_data] env value ARIES_USER will be default username to login aries

wkentaro:
By this PR, you can set `ARIES_USER` to set as default username to login aries.

In my case, `USER` value is `wkentaro` but username in aries is `wada` so I can
tell the Makefile my username by adding a line in my bashrc like below::
```
export ARIES_USER="wada"
```

### DIFF
--- a/jsk_data/Makefile
+++ b/jsk_data/Makefile
@@ -3,7 +3,8 @@
 ##
 all: small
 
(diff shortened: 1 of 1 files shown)


garaemon:
historical reason, we use SSH_USER for this case

wkentaro:
so I should change
`ARIES_USER` -> `SSH_USER` ?

or 
`LOGIN_USER` -> `SSH_USER` ?

garaemon:
ARIES_USER should default to SSH_USER

PR #912, opened 2015-05-05
[image_view2] add depends to gencfg, this breaks deb build

k-okada:


k-okada:
to fix http://jenkins.ros.org/job/ros-hydro-image-view2_binarydeb_raring_amd64/

PR #964, opened 2015-06-02
add file exist check function

aginika:
Check USB device Files

check_file_exists


aginika:
Once close this we will use lsusb to checl USB

PR #965, opened 2015-06-02
add check usb exist with lsusb

aginika:
check_usb_exist


garaemon:
merge origin/master

k-okada:
In hrp case, you have to check if there are 4 devise on the bus

PR #968, opened 2015-06-02
check local remote ROS Parameter diff

aginika:
Very simple script only for check param.
I couldn't find out the way to get the result from mosh, so I use now ssh.

Usage:
```
check_param baxter@localhost /robot_description
```


garaemon:
```python
import subprocess
output = subprocess.check_output(["ssh", "foo", "bash" , ...])
```

aginika:
Thank you, I updated

Issue #919, opened 2015-05-08
Please split out stable packages into separate release units


tfoote:
With all these packages bundled together each release causes a lot of build time on the farm. (On the order of 24hours build time per rosdistro released) In each release very few packages appear to have changes applied to them, but because they are released together all of them are required to rebuild not just the changed packages but also the unchanged packages as well as anything that depends upon them.

I audited the length of time to build each package in this repository.  No individual package is a major problem, it's just that they all add up especially when repeated across the different architectures. 

This is one of the most frequently released packages, which we would like to encourage, however from reviewing our usage releasing the unchanged packages is causing the builds of this repository to consume a large fraction of our buildfarm's resources. 

Here's the details from my analysis. It's based on a manual sampling of the latest trusty amd64 builds for indigo. This analysis does not capture downstream dependency build time.

jsk_tilt_laser takes 12 minutes
speech_recognition_msgs takes 8 minutes
nlopt takes 11 minutes
multi_map_server takes 19 minutes
dynamic_tf_publisher takes 5 minutes
bayesian_belief_networks takes 6 minutes
posedetection_msgs takes 11 minutes
jsk_common takes 5 minutes
jsk_footstep_msgs takes 3 minutes
jsk_tools takes 7 minutes
jsk_topic_tools takes 17 minutes
jsk_network_tools takes 11 minutes
jsk_gui_msgs takes 5 minutes
rostwitter takes 13 minutes
assimp_devel takes 16 minutes
rospatlite takes 8 minutes
voice_text takes 7 minutes
ff takes 12 minutes
ffha takes 8 minutes
jsk_hark_msgs takes 4 minutes
image_view2 takes 17 minutes
jsk_data takes 2 minutes
rosping takes 10 minutes
julius takes 9 minutes
libsiftfast takes 11 minutes
virtual_force_publisher takes 11 minutes
opt_camera takes 15 minutes
mini_maxwell takes 7 minutes
downward takes 15 minutes
sklearn takes 11 minutes
Sum of the time per arch is 296 minutes
On 6 arches that's 29.6 hours
On 2 distros that's 59.2 hours

k-okada:
sorry to make you trouble, We'll fix them:

the list of recent updates:
```
May 9 1.0.70 jsk_data, jsk_footstep_msgs, jsk_netowrk_tools, jsk_tilt_laser, jsk_topic_tools
May 4 1.0.69 image_view2, jsk_tools, jsk_topic_tools,
[...]

PR #988, opened 2015-06-12
split jsk_common into jsk_common_msgs and jsk_3rdparty https://github.com/jsk-ros-pkg/jsk_common/issues/919

k-okada:


k-okada:
*ATTENTION*

This will split jsk_common into
jsk_common, jsk_common_msgs and jsk_3rdparty

k-okada:
Fixed #919

PR #990, opened 2015-06-14
[jsk.rosbuild] Add --no-sudo option to run without sudo

garaemon:


garaemon:
This should be good for install ros codes on pr2 user directoris.

PR #995, opened 2015-06-15
[jsk.rosinstall.hydro] Add catkin to install from source

wkentaro:
To avoid Bug: https://github.com/ros/catkin/issues/686
Issue: https://github.com/jsk-ros-pkg/jsk_common/issues/885

### DIFF
--- a/jsk.rosinstall.hydro
+++ b/jsk.rosinstall.hydro
@@ -10,3 +10,7 @@
     local-name: image_pipeline
     uri: https://github.com/ros-perception/image_pipeline.git
     version: hydro-devel
+- git:
+    local-name: catkin
+    uri: https://github.com/ros/catkin.git
+    version: indigo-devel


wkentaro:
catkin does not have `hydro-devel` branch.

k-okada:
Hum

Who need to use zsh on hydro?

wkentaro:
I don't really know about each person's shell preference.
But the problem is also described in https://github.com/jsk-ros-pkg/trans_system/pull/238 .

wkentaro:
And I do use zsh on hydro.

k-okada:
I think you do not have to use hydro.

2015年6月16日火曜日、Kentaro Wada<notifications@github.com>さんは書きました:

> And I do use zsh on hydro.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jsk-ros-pkg/jsk_common/pull/995#issuecomment-112150540>
> .
>


-- 
--
◉ Kei Okada

garaemon:
I'm using zsh on hydro and I have no problem though

garaemon:
Anyway, I think this commit has no problem for us.

k-okada:
no problem at this moment, but I'm afraid this will have some trouble in
future.

If we can assume new version of catkin within a few weeks/month, we will
revert this PR. but most user did not realize that and they did not know
how to handle such case, then they will still use old version of catkin or
latest version of catkin. If this PR is required only for two super expert
developer, we'll better to keep this as PR and ask them to merge by them
self or manually install latest version of catkin.


--
◉ Kei Okada

wkentaro:
I get your point.
It seems that we should avoid to add `ros/catkin` to the common `.rosinstall`.

PR #1005, opened 2015-06-18
[jsk.rosbuild] fix no sudo option in jsk.rosbuild

furushchev:
in existing function `runsudo` used `sudo sh -c "$*"`, which collapse some code that includes redirects.
for example:

```
runsudo sh -c "echo \"deb $DEBIAN_REPOSITORY `lsb_release -cs` main\" > /etc/apt/sources.list.d/ros-latest.list"
```

this is expanded to:

```
sudo sh -c 'sh -c echo "deb http://packages.ros.org/ros-shadow-fixed/ubuntu trusty main" > /etc/apt/sources.list.d/ros-latest.list'
```

as above, expanding codes using `"$*"` removes top-level `"`, which causes unexpected behavior.

### DIFF
--- a/jsk.rosbuild
+++ b/jsk.rosbuild
@@ -30,6 +30,7 @@ if [ $? != 0 ]; then
     usage
 fi
 DEBIAN_REPOSITORY='http://packages.ros.org/ros/ubuntu'
+SUDO=sudo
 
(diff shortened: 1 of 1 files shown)


furushchev:
I see Thanks.

PR #1006, opened 2015-06-18
- pseudo jacobian has different for m>n and n<m

k-okada:


k-okada:
@kochigami this may solve you problem, we found this bug with @mmurooka see https://ja.wikipedia.org/wiki/%E6%93%AC%E4%BC%BC%E9%80%86%E8%A1%8C%E5%88%97#.E6.80.A7.E8.B3.AA or ask him about detail

Also I found wrist roll joint of Pepper robot is too week and always have large effort (errors) so using
`  <arg name="tip"  default="LForeArm"/>` instead of `l_wrist` may good. Another option is to use `  <arg name="root" default="Tibia"/>` instead of `torso`, but not sure if this really gets good results or not

kochigami:
Sorry for my late reply.
I appreciate your advice.

I tried the program with 
1. using ```LForeArm``` instead of ```l_wrist``` and ```Tibia``` instead of ```torso```.
case of joint link dof (7) >  end effector dof (6)
(kdl_chain(0) KneePitch, kdl_chain(1) HipPitch, kdl_chain(2) HipRoll, kdl_chain(3) LShoulderPitch, kdl_chain(4) LShoulderRoll, kdl_chain(5) LElbowYaw, kdl_chain(6) LElbowRoll)

2. using ```LForeArm``` instead of ```l_wrist``` and ```torso```.
case of joint link dof (4) < end effector dof (6)
(kdl_chain(0) LShoulderPitch, kdl_chain(1) LShoulderRoll, kdl_chain(2) LElbowYaw, kdl_chain(3) LElbowRoll)

Results:
Both->
The arrow movement is very stable.
In addition to this, I found the direction of force arrow indicates the internal power which is exerted to pepper when I move pepper's arm. 

Difference->
In case of 1, the arrow can indicate a horizontal direction (right or left), but in case of 2, it only indicates front or back direction even if I pulled pepper's arm in a horizontal direction.
I found the redundancy of joint link dof is important. Therefore, using ```Tibia``` is better.

PR #1034, opened 2015-07-16
[jsk_tools/99.jsk_tools.sh] Safer rost and rosn func and also supports rosmsg show

wkentaro:
rost_rosn


garaemon:

PR #1037, opened 2015-07-17
jsk.rosbulid: add -r option to specify package to download

k-okada:
```
./jsk.rosbuild -y -s --rtm --from-source -r hrpsys,rtm-ros-robotics/rtmros_common indigo
```
download only hrpsys and rtmros_common, so they do not have to compile everything written in jsk.rosinstall

### DIFF
--- a/jsk.rosbuild
+++ b/jsk.rosbuild
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 function usage {
-    echo >&2 "usage: $0"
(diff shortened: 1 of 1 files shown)


k-okada:
https://github.com/start-jsk/rtmros_common/pull/773this depends on https://github.com/jsk-ros-pkg/jsk_roseus/pull/342 and

k-okada:
http://unix.stackexchange.com/questions/22820/how-to-make-apt-get-accept-new-config-files-in-an-unattended-install-of-debian-f

Issue #1038, opened 2015-07-20
sometimes ConnectionBasedNodelet failed to detect connection/disconnection


garaemon:
from https://github.com/jsk-ros-pkg/jsk_demos/issues/998

1. bond problem
  
  (solved by https://github.com/jsk-ros-pkg/jsk_common/pull/1041)
  see [this gist](https://gist.github.com/garaemon/0c1370e7b8a2db33f20c) to reproduce it
2. timing issue between `ConnectionBasedNodelet::connectionCallback` and `ConnectionBasedNodelet::advertise` (solved by https://github.com/jsk-ros-pkg/jsk_common/pull/1040)

  see [this gist](https://gist.github.com/garaemon/a67e025f1018f492c1eb) to reproduce it

garaemon:
It's not a test of ConnectionBasedNodelet, but a test of `jsk_topic_tools/Relay`.
https://gist.github.com/garaemon/f2a8feea5a4b2193641c

It's a simple test to chain 100 relays.

according to this test, it takes 20 seconds to get output from `rostopic echo relay_99/output`.

garaemon:
I found that one of the reasons to kill nodelet is bond mechanism.
nodelet manager keeps publishing "~bond" topic in about 20 Hz.
if nodelet loaders misses to listen this topic, the loaders exits. ([here](https://github.com/ros/nodelet_core/blob/indigo-devel/nodelet/src/nodelet.cpp#L337))

Default timeout is 2-10 secs [see here](https://github.com/ros/bond_core/blob/master/bond%2Fmsg%2FConstants.msg)

garaemon:
And if bond is broken, loader unload the nodelet https://github.com/ros/nodelet_core/blob/indigo-devel/nodelet/src/nodelet.cpp#L345

garaemon:
I found a race condition in `jsk_topic_tools::ConnectionBasedNodelet::advertise`.

If a topic is subscribed right after advertise it, connection callback is called in [here](https://github.com/jsk-ros-pkg/jsk_common/blob/d947028a16ace7d466791d7ebe5e783bbee12e73/jsk_topic_tools/include/jsk_topic_tools/connection_based_nodelet.h#L126) and it is before the publisher is added to `publishers_`.

[...]

PR #1039, opened 2015-07-20
[jsk_topic_tools] Add StringRelay nodelet to test DiagnosticNodelet class

garaemon:
To test for https://github.com/jsk-ros-pkg/jsk_common/issues/1038

* Do not add to unittest. because it can easily fail on travis.


garaemon:

PR #1040, opened 2015-07-20
[jsk_topic_tools] Make advertise template method critical section in order to avoid race condition between advertise and connectionCallback

garaemon:
see https://github.com/jsk-ros-pkg/jsk_common/issues/1038


garaemon:

PR #1050, opened 2015-07-27
[jsk_network_tools] Kill child process at the end of process in silve…

garaemon:
…rhammer_highspeed_receiver.py

@furushchev is it correct?

### DIFF
--- a/jsk_network_tools/scripts/silverhammer_highspeed_receiver.py
+++ b/jsk_network_tools/scripts/silverhammer_highspeed_receiver.py
@@ -244,5 +244,7 @@ class SilverHammerReceiver:
 if __name__ == "__main__":
     rospy.init_node("silverhammer_highspeed_receiver")
     receiver = SilverHammerReceiver()
-    rospy.on_shutdown(receiver.on_shutdown)
-    receiver.run()
+    try:
+        receiver.run()
+    finally:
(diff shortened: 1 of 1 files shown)


garaemon:
ping @furushchev

furushchev:
@garaemon Sorry for late response.
I could not understand what you want to change.
we are already checking if we need to kill child process by checking at https://github.com/garaemon/jsk_common/blob/89fdb1358e34fedaa368eaaeefa06d1f01ab8f74/jsk_network_tools/scripts/silverhammer_highspeed_receiver.py#L128
Is there any problem?

garaemon:
> Is there any problem?

Yes sometimes child process would be a zombie.

garaemon:
I see, `rospy.on_shutdown` is just a registration of hook function.

However, it will not be called if a process is killed by SIGTERM, which is fired by roslaunch.

furushchev:
@garaemon Thank you very much for clarifying! Now I think it's ok to merge!

garaemon:
finally section does not run if SIGTERM is fired.

garaemon:
Is it difficult to implement without multiprocessing?

two ideas:
1) Use cython
2) Use c++ to concatenate packets (use roscpp?)

PR #1061, opened 2015-08-03
[jsk_data] Add more topics to record in hrp2_rosbag_always.sh

garaemon:


garaemon:
merged because it fails due to DNS problem

PR #1062, opened 2015-08-03
[jsk_data/rosbag_always.py] Supress message about directory size and

garaemon:
colorize message about removing bag files


garaemon:
merged because it fails due to DNS problem

PR #1064, opened 2015-08-04
[jsk_data] Popup notification on desktop when removing a bag file

garaemon:


garaemon:
It should be like this

screenshot_from_2015-08-04 17 10 23

garaemon:

PR #1065, opened 2015-08-04
[jsk_tools] Record /urata_status topic in hrp2_rosbag_always.sh

garaemon:


garaemon:

PR #1066, opened 2015-08-04
[jsk_network_tools] Add C++ implementation of silverhammer_highspeed_receiver.

garaemon:
Only core is implemented in c++.

It cooperates with python thin wrapper and bridged via ros between them.

* python is too slow to implement receiver in one process
  * Giant lock does not allow us to write parallel processes
* python multiprocessing may generate zombie process
  * Because python signal/exception mechanism is not robust with multiprocessing and socket programming
* Implement UDP server in C++ to write `receiving packets` and `collecting packets` in different thread for performance reason.
* But C++ is not good at Meta Object Protocol, so C++ UDP server just sends not-deserialized ROS buffer to python wrapper via normal ROS topic protocol and convert it into serialized ROS message in python, because we'd like to specify different classes from ros parameters.

### DIFF
--- a/jsk_network_tools/CMakeLists.txt
+++ b/jsk_network_tools/CMakeLists.txt
@@ -3,7 +3,7 @@ project(jsk_network_tools)
 
 find_package(catkin REQUIRED COMPONENTS
   message_generation std_msgs sensor_msgs rostest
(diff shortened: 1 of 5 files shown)


garaemon:

PR #1121, opened 2015-09-11
Remove groovy and add indigo on REAME.md

garaemon:

PR #1122, opened 2015-09-11
Rename .rosinstall -> .travis.rosinstall

wkentaro:


wkentaro:
duplicates with #1121

PR #1130, opened 2015-09-17
Add document how to install scudcloud on 12.04

garaemon:
scudcloud is a desktop client of slack.
You can use it on 14.04 without no effort but on 12.04 we need ppas.

screenshot from 2015-09-17 18 03 37


garaemon:
こういう開発に便利なツールtipsは別にまとめたほうがいいかも

PR #1154, opened 2015-10-01
[jsk_data] Add file completion in bash

wkentaro:


wkentaro:
image

Actually, this does not work with bashcompinit. ( lots of no match error )
so we might need to write another zsh completion script.

wkentaro:
so I remove zsh completion at this time.

PR #1157, opened 2015-10-03
[jsk_topic_tools] Ros error for rosparam type conversion

wkentaro:
For: https://github.com/jsk-ros-pkg/jsk_recognition/issues/1188

### DIFF
--- a/jsk_topic_tools/src/rosparam_utils.cpp
+++ b/jsk_topic_tools/src/rosparam_utils.cpp
@@ -34,6 +34,7 @@
  *********************************************************************/
 
 #include "jsk_topic_tools/rosparam_utils.h"
+#include "jsk_topic_tools/log_utils.h"
 
 namespace jsk_topic_tools
 {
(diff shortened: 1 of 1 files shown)


jsk_topic_tools/src/rosparam_utils.cpp
```diff
@@ -51,6 +52,7 @@ namespace jsk_topic_tools
     }
     default:
     {
+      JSK_ROS_ERROR("the value cannot be converted into double");
```
garaemon: I think `XmlRpc::XmlRpcValue` can be printed by `<<` operator. Can you output it?

garaemon: like
```c++
ROS_ERROR_STREAM("the value cannot be converted into double: " << val);
```

wkentaro:
image

I updated the commit.

wkentaro:
```
+ git clone -q https://github.com/ros/catkin

error: RPC failed; result=56, HTTP code = 200

fatal: The remote end hung up unexpectedly

fatal: early EOF

fatal: index-pack failed

++ error
```

garaemon:
may be rejected by github because of too much connection

Issue #1141, opened 2015-09-29
[enhancement] [jsk_data] Select file to install interactively


garaemon:
I'd like to install file interactively when I'm using jsk_data.
one useful library is [python-inquirer](https://github.com/magmax/python-inquirer)

wkentaro:
:+1: 

we can also use percol::
![out](https://cloud.githubusercontent.com/assets/4310419/10267363/543dcbd4-6acc-11e5-9c62-d93876b416dd.gif)

garaemon:
Can you invoke percole in `jsk_data get` and select file by percole and download it?

PR #1159, opened 2015-10-04
[jsk_data] Select filename interactively with percol in pubinfo

wkentaro:
see: https://github.com/jsk-ros-pkg/jsk_common/issues/1141#issuecomment-145332290


garaemon:
:+1:

PR #1165, opened 2015-10-08
Hide warning output from lsof because of user problem.

snozawa:
When `sudo lsof -l` is called, 
```
lsof: WARNING: can't stat() fuse.gvfs-fuse-daemon file system /home/hrpuser/.gvfs
      Output information may be incomplete.
```
occurs.

This is because root user cannot access other user fuse system:
http://unix.stackexchange.com/questions/171519/lsof-warning-cant-stat-fuse-gvfsd-fuse-file-system

One solution is set `allow_root` in /etc/fuse.conf, but it seems to have too large influences.

So I hide this warning message by redirecting.

@garaemon , how about this?

### DIFF
--- a/jsk_tools/src/jsk_tools/sanity_lib.py
+++ b/jsk_tools/src/jsk_tools/sanity_lib.py
@@ -247,7 +247,7 @@ def checkUSBExist(vendor_id, product_id, expect_usb_nums = 1, host="", success_m
 def getROSMasterCLOSE_WAIT(host, username=""):
     if username != "":
         host = username + "@" + host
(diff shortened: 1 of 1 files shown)


garaemon:
LGTM

Another solution:
https://github.com/jsk-ros-pkg/trans_system/issues/432

snozawa:
I think jsk-ros-pkg/trans_system#432 is more reasonable solution.
I'll check tomorrow.
Thanks.

garaemon:
> I think jsk-ros-pkg/trans_system#432 is more reasonable solution.

I don't think so. In this script, we can ignore stderr output.

But the source of this problem is gvfs and I don't know why gvfs is enabled. I believe gvfs is disable in normal ubuntu.

snozawa:
>    I think jsk-ros-pkg/trans_system#432 is more reasonable solution.

I checked jsk-ros-pkg/trans_system#432 solves this PR's problem.

PR #1178, opened 2015-10-12
[jsk_topic_tools] Python ConnectionBasedTransport

wkentaro:


wkentaro:
This PR contains:

* rostest for cpp ConnectionBasedNodelet
* implementation of python ConnectionBasedTransport
* rostest for python ConnectionBasedTransport

jsk_topic_tools/CMakeLists.txt
```diff
@@ -98,6 +98,8 @@ if (NOT $ENV{ROS_DISTRO} STREQUAL "indigo")
   add_rostest(test/test_hz_measure.test)
   add_rostest(test/test_block.test)
 endif(NOT $ENV{ROS_DISTRO} STREQUAL "indigo")
```
wkentaro: In fact, I don't know why there is condition about indigo here..

garaemon: older version of catkin could not execute `add_rostest`

garaemon:
I think this feature should be implemented in jsk_recognition_utils

wkentaro:
So the location of connectionbasednodelet is also wrong?

garaemon:
> connectionbasednodelet

It does not expect `sensor_msgs/Image`.

jsk_topic_tools/CMakeLists.txt
wkentaro: So you mean the catkin version on `hydro` is newer than that on `indigo`?
I wonder why `NOT ... STREQUAL`.

garaemon: newer than default hydro but older than current latest indigo.

It should be fixed the latest indigo catkin.

wkentaro:
ConnectionBasedTransport does not expect `sensor_msgs/Image` either.
I used `sensor_msgs/Image` in the sample.

garaemon:
> I used sensor_msgs/Image in the sample.

I see

jsk_topic_tools/CMakeLists.txt
garaemon: see https://github.com/ros/catkin/issues/703

wkentaro: I see.
So at this time, should I write `add_rostest` lines in side the if condition?

wkentaro: I updated the commit.

jsk_topic_tools/src/jsk_topic_tools/transport.py
```diff
@@ -0,0 +1,55 @@
...
+import rospy
+
+
+ConnectionStatus = Enum('ConnectionStatus', 'SUBSCRIBED NOT_SUBSCRIBED')
+
+
+class ConnectionBasedTransport(rospy.SubscribeListener):
+    def __init__(self, node_name):
+        super(ConnectionBasedTransport, self).__init__()
+        rospy.init_node(node_name)
```
garaemon: I think programmer should call `rospy.init_node` before creating ConnectionBasedTransport.
It would be problem if user want to create several instances of ConnectionBasedTransport

wkentaro:
I updated the commit about `add_rostest` in CMakeLists.txt.

garaemon:
> I updated the commit about add_rostest in CMakeLists.txt.

uncomment if-block after this PR is merged.

jsk_topic_tools/src/jsk_topic_tools/transport.py
wkentaro: Well spotted thanks. I will change it.

wkentaro:
> uncomment if-block after this PR is merged.

Which lines comment should I remove?

jsk_topic_tools/src/jsk_topic_tools/transport.py
wkentaro: > I think programmer should call rospy.init_node before creating ConnectionBasedTransport.
It would be problem if user want to create several instances of ConnectionBasedTransport

I updated the commit.

garaemon:
> Which lines comment should I remove?

https://github.com/jsk-ros-pkg/jsk_common/blob/master/jsk_topic_tools/CMakeLists.txt#L90

We may have trouble in running test like https://github.com/jsk-ros-pkg/jsk_common/pull/1177

wkentaro:
> https://github.com/jsk-ros-pkg/jsk_common/blob/master/jsk_topic_tools/CMakeLists.txt#L90

So you mean comment out the `if()` and `endif()` ? Because the bug has been fixed in the latest catkin.

> We may have trouble in running test like #1177

like #177 ? You mean a bug of catkin?

garaemon:
I'm just afraid that these tests are not stable.
I don't want to add unstable test code and PR should be separated to see what will happen in testing.

wkentaro:
I see.
And do you mean I should split the PR?

2015年10月12日月曜日、Ryohei Ueda<notifications@github.com>さんは書きました:

> I'm just afraid that these tests are not stable.
> I don't want to add unstable test code and PR should be separated to see
> what will happen in testing.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jsk-ros-pkg/jsk_common/pull/1178#issuecomment-147403528>
> .
>


-- 
和田 健太郎 / Kentaro Wada
http://wkentaro.com

garaemon:
If it raises strange error, split PR.
If not, you don'nt need to do

wkentaro:
python-enum34 is not released on apt for precise.
https://github.com/ros/rosdistro/pull/9635

PR #1218, opened 2015-11-09
[image_view2] Publish cropped image with roi

wkentaro:
Currently this is only work with rectangle mode.


wkentaro:
image_view2_screenrectangle_image

PR #1219, opened 2015-11-09
[image_view2] Visualize depth image

wkentaro:
*before*
image

*after*
image_view2_view_depth_image


wkentaro:
for image_view: https://github.com/ros-perception/image_pipeline/pull/157

PR #1224, opened 2015-11-10
[jsk_topic_tools] Function to warn with no remappings

wkentaro:


k-okada:
this is general comment on topic tools

- using python on this type of topic tools may not a good idea, especially
for simulated environment; they have significant performance issues
discussed at
http://answers.ros.org/question/29425/rospy-horrible-performance-with-sim_time/
(btw, kwc:ken is the original rospy coder, JSK mentor during pr2 program,
 and moved to G 1-2 year before they acquired many robotics company.), so
if we use this kind of highspeed message relay tools, we'd better to choose
c++ client (or roseus which call c++ function internally), or dig into the
python implementation and find the solution. of course, this is the lowest
priority task.
Cf.
https://github.com/start-jsk/rtmros_common/issues/692#issuecomment-94121971

--
◉ Kei Okada

wkentaro:
> -  using python on this type of topic tools may not a good idea, especially
for simulated environment; they have significant performance issues
discussed at

Where is bottleneck? [here](https://github.com/jsk-ros-pkg/jsk_common/pull/1224/files#diff-5c8c09b0d5933af83695ca5f9728ab7bR43)?
> if we use this kind of highspeed message relay tools

I think this is not relay tool precisely, do you say 'relay' as example?

wkentaro:
> - using python on this type of topic tools may not a good idea, especially for simulated environment; they have significant performance issues discussed at

and why do you specify the situation `use_sim_tim:=true` ?
because each node has to listen `/clock` topic?

wkentaro:
```
% echo "$(rosnode list | wc -l) , $(rostopic list | wc -l)"
94 , 495
[http://baxter:11311][133.11.216.160] wkentaro at anaconda in ~ 
% ./warn_no_remap_performance.py
rospy.get_name
elapsed: 2e-06 [s]
rospy.names.resolve_name
elapsed: 1.1e-05 [s]
['/spam/input', '/spam/input/a', '/spam/output', '/spam/output/a']
rospy.names.get_resolved_mappings
elapsed: 1e-06 [s]
{}
```
[warn_no_remap.py](https://gist.github.com/3a65157a1f378206272e)

wkentaro:
> if we use this kind of highspeed message relay tools, we'd better to choose
c++ client (or roseus which call c++ function internally), or dig into the
python implementation and find the solution. of course, this is the lowest
priority task.

This is a tool when writing node of python, so we cannot use c++ or roseus as the substitution.
So if the performance is too critical, I need change rospy implementation. (maybe call c++ function inside it if currently not)

k-okada:
> do you say 'relay' as example?

yes, and he said listing topic running at 1000hz may consume CPU power, 
which may not become a problem on C++ implementation. so I believe you want 
to write everything in python, but  you may need to fix this problem at 
some point, not now.

--
◉ Kei Okada

PR #1229, opened 2015-11-11
[jsk_topic_tools] warnNoRemap for cpp nodes

wkentaro:


wkentaro:
This is cpp version of #1224 
image

k-okada:
nice

Issue #1145, opened 2015-09-29
[fixme] [jsk_topic_tools] Enable 'always_subscribe' without another method call


wkentaro:
The method named onInitPostProcss. 
At present, it need to be called in subclasses of ConnectionBasedNodelet.

Discussed at https://github.com/jsk-ros-pkg/jsk_common/pull/1140/files#r40654022

PR #1231, opened 2015-11-11
[jsk_topic_tools] Show input/output topics with --inout opt

wkentaro:


wkentaro:
image

wkentaro:
This depends on https://github.com/jsk-ros-pkg/jsk_common/issues/1145 to make this work for all nodes' subscriptions.
But does work for publications.

PR #1261, opened 2015-12-03
[jsk_tools] Replace image of topic_hz_monitor

wkentaro:
The command in the image was wrong in previous version.


garaemon:
texttable is not resolved on saucy, failed to release 2.0.7

wkentaro:
Ok I will add saucy and utopic. 
Do you know how can I check whether the apt package is released?
They are minor version and I could not find them at packages.ubuntu.com.

wkentaro:
But I'm wondering why we need to support saucy and utopic.

PR #1276, opened 2015-12-07
[jsk_topic_tools/log_utils] Fix include guard.

garaemon:
Define warnNoRemap in include guard section.


garaemon:
Will release new version after it is merged

PR #1281, opened 2015-12-09
[image_view2] avoid SEGV caused by negative width for rectangle

YuOhara:
for https://github.com/jsk-ros-pkg/jsk_common/issues/1247


wkentaro:
Thanks.

PR #1283, opened 2015-12-09
[jsk_tools] Add tool to make commit message informative

wkentaro:
Closes #1202

This is proposed by @k-okada and discussed on #1202

Modified:
	 jsk_tools/env-hooks/99.jsk_tools.sh

### DIFF
--- a/jsk_tools/env-hooks/99.jsk_tools.sh
+++ b/jsk_tools/env-hooks/99.jsk_tools.sh
@@ -206,3 +206,21 @@ rosrecord () {
     return 1
   fi
 }
+
+
+_jsk_tools_what_ros_pkg () {
+  looking_path=$(pwd)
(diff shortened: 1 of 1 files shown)


wkentaro:
Please run `jsk_git_commit_verbose` when committing.

garaemon:
If you rename the function like `git-foo-bar`, is it possible to use like `git foo-bar`?

wkentaro:
I think we need to create an executable which is named `git-foo-bar` to do that.

garaemon:
I see

wkentaro:
> If you rename the function like git-foo-bar, is it possible to use like git foo-bar?

Do you think it's better?

garaemon:
> Do you think it's better?

I think so.

wkentaro:
```
git jsk-commit
```
is ok?

PR #1291, opened 2015-12-10
[image_view2] Test screenrectangle image with mouse event

wkentaro:
For https://github.com/jsk-ros-pkg/jsk_common/issues/1247#issuecomment-163701214

Modified:
	 jsk_ros_patch/image_view2/package.xml
Added:
	 jsk_ros_patch/image_view2/test/publish_lena.py
	 jsk_ros_patch/image_view2/test/publish_mouse_event.py
	 jsk_ros_patch/image_view2/test/test_image_view2.test

PR #1303, opened 2015-12-15
[image_view2] retry 3 to avoid 'failed to launch' error

wkentaro:


wkentaro:
for https://github.com/jsk-ros-pkg/jsk_common/pull/1291#issuecomment-164495447

PR #1307, opened 2015-12-15
Update travis for pip

garaemon:
It includes #1303, #1305, #1306


garaemon:
https://github.com/jsk-ros-pkg/jsk_common/pull/1303 is not enough.
I have seen the same error

PR #1324, opened 2015-12-24
[jsk_data] Diretly use drive command locally for faster uploading

wkentaro:


wkentaro:
This simplifies things. 
jsk_data uploaded files to both aries and google drive.
After this pr, user need to configure gdrive in local at once, but does not need to upload both but only specified location.
(private -> aries, public -> google drive)

PR #1402, opened 2016-06-05
add parameter for selecting MultiThread callback or SingleThread callback

YoheiKakiuchi:


k-okada:
what is the difference between these two ?
need doc? http://jsk-docs.readthedocs.io/en/latest/jsk_common/doc/jsk_topic_tools/index.html

YoheiKakiuchi:
> what is the difference between these two ?

MT node handle generate threads for callback function.
It is good for nodelet that have to process many callback function.
But, heavy callback function cause to generate many thread for high rate topic.
It spend many CPU cores.

PR #1418, opened 2016-06-21
Enhance data_collection_server regarding of response, params, and sample

wkentaro:
- Add sample for data_collection_server in jsk_data
- Return saved message as TriggerResponse in data_collection_server
- Make params as optional for data_collection_server
- Change dynamically save_dir parameter in data_collection_server


wkentaro:
Please merge this.

PR #1421, opened 2016-06-22
[jsk_topic_tools] Add JSK_ROS_XXX_THROTTLE, JSK_ROS_XXX_STREAM_THROTTLE

wkentaro:


k-okada:
Please add link to Pull Request in upstream package

wkentaro:
This change maybe only for jsk_topic_tools users.

k-okada:
You can improve normal `ROS_...STREAM... ` so we do not need to define this macro at here.

PR #1422, opened 2016-06-26
Fix bug for new savetype YAML in data_collection_server.py

wkentaro:


k-okada:
Progress Bar error , need update jsk_travis?

PR #1434, opened 2016-08-07
[jsk_data] use urdf model with hand for robot_description when playing with hrp2

mmurooka:
HRP2 model with HRP3 hand is used as default.
By setting `use_collada_model:=true`, you can use model in previous version, (but I think no one wants to use model without hand.)


mmurooka:
all tests have passed. please merge if there is no problem.

PR #1436, opened 2016-08-08
[jsk_tools/env-hooks/99.jsk_tools.sh] use ip for rossetip instead of ifconfig

furushchev:
for future compatibility and multi platform.

### DIFF
--- a/jsk_tools/env-hooks/99.jsk_tools.sh
+++ b/jsk_tools/env-hooks/99.jsk_tools.sh
@@ -64,6 +64,7 @@ rossetmaster() {
     fi
     echo -e "\e[1;31mset ROS_MASTER_URI to $ROS_MASTER_URI\e[m"
 }
+
 rossetrobot() {
     echo -e "\e[1;31m *** rossetrobot is obsoleted, use rossetmaster ***\e[m"
     rossetmaster $@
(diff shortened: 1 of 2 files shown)


furushchev:
@k-okada Test passed. Please merge.

k-okada:
1) ip is provided from iproute2 package, did you add rosdep for this?
```
$ dpkg -S /bin/ip
iproute2: /bin/ip
```

2) do we have test code ros rossetip?

3) is this just a refactor ? or you have ane specific reason to change code?

furushchev:
@k-okada Thank you for reviewing.

> 1) ip is provided from iproute2 package, did you add rosdep for this?

No, just forgot. I'll add it.

> 2) do we have test code ros rossetip?
 
No, I think. I should add test code. (but a bit difficult)

> 3) is this just a refactor ? or you have ane specific reason to change code?

current implementation uses `ifconfig`. this is deprecated and will be removed in the future. (already unavailable on OSX), and replaced to `ip`.

Also `ip` provides function to get route from specific device to host and it's much smarter to use this instead of masking manually. (https://github.com/jsk-ros-pkg/jsk_common/pull/1436/files#diff-48005ba849f01087afb5a38e0561c822L97)

Also I fixed weird behavior in `rossetip_addr`, which captures ip address of `docker0` when docker is installed.

furushchev:
@k-okada How about this pull request?

k-okada:
Did you ass iproute2 to package.xml ?

furushchev:
@k-okada Oh,,, very sorry... added.

k-okada:
memo:
- users do not install from scratch, they just update the code
- when users have trouble, they roll back code to previous version
https://github.com/jsk-ros-pkg/jsk_common/issues/1472#issuecomment-256624899

PR #1446, opened 2016-08-26
[jsk_topic_tools] add JSK_NODELET_LOG_THROTTLE

furushchev:


jsk_topic_tools/CMakeLists.txt
```diff
@@ -125,6 +125,16 @@ if (CATKIN_ENABLE_TESTING)
...
+
+  add_library(jsk_topic_tools_test
+    src/tests/test_nodelet_log_utils.cpp
+    src/log_utils.cpp)
+  target_link_libraries(jsk_topic_tools_test
+    ${catkin_LIBRARIES} ${BOOST_LIBRARIES})
+  add_executable(test_nodelet_log EXCLUDE_FROM_ALL test/test_nodelet_log.cpp)
+  target_link_libraries(test_nodelet_log ${catkin_LIBRARIES} ${GTEST_LIBRARIES})
+  add_dependencies(tests test_nodelet_log)
+  add_rostest(test/test_nodelet_log.test)
```
wkentaro: Is `add_rostest_gtest()` not usable?

furushchev: @wkentaro I did not know just how to test by `add_rostest_gtest`.

wkentaro:
Actually, I found the `JSK_ROS_XXX` logging macros are not necessary just recently.
Its feature is already covered by `ROSCONSOLE_FORMAT` environmental variable.
http://wiki.ros.org/rosconsole#Console_Output_Formatting
(you can try it by running `export ROSCONSOLE_FORMAT='[${node}] [${function}] [${severity}] [${time}]: ${message}'`)

I agree that adding this function to jsk_topic_tools at this time, but we should add deprecation warning for all of the logging functions.
like
```
Deprecation warning: JSK_ROS_XXX is being deprecated, please consider customize rosconsole output by using ROSCONSOLE_FORMAT environmental variable. See http://wiki.ros.org/rosconsole#Console_Output_Formatting for more details.
```

jsk_topic_tools/CMakeLists.txt
wkentaro: I suppose you can do like this:
https://github.com/jsk-ros-pkg/jsk_common/blob/master/jsk_topic_tools/CMakeLists.txt#L126-L127

k-okada:
> I agree that we add this function to jsk_topic_tools at this time, but we should add deprecation warning for all of the logging functions.

why we have to add this function at this time, for hydro?

wkentaro:
@k-okada It is not necessary. But currently there are already logging functions of `JSK_ROS_XXX`, `JSK_ROS_XXX_STREAM`, `JSK_ROS_XXX_THROTTLE`, `JSK_ROS_XXX_STREAM_THROTTLE`, `JSK_NODELET_XXX`, and `JSK_NODELET_XXX_STREAM`, so I thought some users expect `JSK_NODELET_XXX_THROTTLE` and `JSK_NODELET_XXX_STREAM_THROTTLE` are also supported, and it is better that we add deprecation warning for them at this time and will remove all macros at once in the next major release. What do you think?

k-okada:
I see, Since we already noticed that JSK_XXX logging macro is no longer needed, so we can create PR that display deprecation warning for these macros

PR #1454, opened 2016-09-06
[jsk_topic_tools] Fix missing installation of jsk_topic_tools_test_nodelet.xml

wkentaro:
This is critical because the xml file is declared in package.xml.
If I install jsk_topic_tools with apt, launch file with nodelet
won't work because of not found error.
Could you please merge this and release next version, soon?


k-okada:
Ok released https://github.com/ros/rosdistro/pull/12626

wkentaro:
Thank you!

PR #1457, opened 2016-10-06
[jsk_topic_tools] set rate with param in tf_to_pose.py

mmurooka:


mmurooka:
Travis test passed. Cloud you merge this?

k-okada:
1, why you need to convert tf to pose, why client program directly listens tf ?
2. why you publish pose periodically , why they do not publish pose only when new tf arrives.

k-okada:
and why you publish pose with current time stamp, not tf time stamp?

mmurooka:
> 1, why you need to convert tf to pose, why client program directly listens tf ?

I'd like to visualize in Rviz as follows. tf is a little bit difficult to manage this because new frame is automatically visualized in Rviz even if we select only target frame.

screenshot from 2016-10-07 10 49 50

>  2. why you publish pose periodically , why they do not publish pose only when new tf arrives.

I'm using vive node or ar marker node, and they publish tf but do not publish pose.

> and why you publish pose with current time stamp, not tf time stamp?

Maybe, this should be fixed, and I'll do this (cc original author: @wkentaro )

mmurooka:
>>    why you publish pose periodically , why they do not publish pose only when new tf arrives.
> I'm using vive node or ar marker node, and they publish tf but do not publish pose.

Sorry, I misunderstood. Maybe this also should be fixed(?).

k-okada:
ok, i understand, if this is  only for visualization, we do not have to be serious on time stamp, but if someone using this for vision-action loop, that's might be dangerous.

mmurooka:
I see. Thank you.

k-okada:
1. as for cart position, you'd better to publish euslisp cart model with "Marker" 
2. although that is not essential, putting AR marker is really bad impression, specially if you use them more than a few years.
2 - a: can we take IR image without loosing performance of depth image? if so we can create marker which is invisible to human eye.
2 - b can we use AR tag, that does not look like AR marker tag? something like ...
http://co-jin.net/wp-content/uploads/2014/03/inferno-sauce.jpg
http://stat.ameba.jp/user_images/20130705/04/korean-ikemen/c0/63/j/o0256028112598858234.jpg
http://livedoor.blogimg.jp/kadench/imgs/1/a/1ad3ed5e.jpg
http://kuma46.com/koguma/wp-content/uploads/2014/06/thumb7.jpg
http://inulab.com/wp-content/uploads/2012/04/30-artistic-creative-barcode-designs.jpg
http://bitwave.showcase-tv.com/wp-content/uploads/2015/04/image221.png

k-okada:
how about this ->
http://4.bp.blogspot.com/_cHodmotLOsw/S3_1azCXToI/AAAAAAAABe0/ggCvuHuKE3E/s320/P1040663.JPG
use use cart and robot instead of camel... ->
http://img-cdn.jg.jugem.jp/c32/1580862/20130823_715555.jpg
we can also check ->
https://kinarino.jp/cat7-%E3%82%A2%E3%83%BC%E3%83%88%E3%83%BB%E3%82%AB%E3%83%AB%E3%83%81%E3%83%A3%E3%83%BC/13891-%E3%81%AA%E3%81%8C%E3%83%BC%E3%81%84%E6%AD%B4%E5%8F%B2%E3%81%8C%E8%82%B2%E3%81%A6%E3%81%9F%E3%83%87%E3%82%B6%E3%82%A4%E3%83%B3%EF%BC%81%E3%82%A4%E3%82%B9%E3%83%A9%E3%83%A0%E6%A8%A1%E6%A7%98%E3%80%81%E3%81%9D%E3%81%AE%E9%AD%85%E6%83%91%E3%81%AE%E4%B8%96%E7%95%8C


--
◉ Kei Okada


2016-10-07 11:05 GMT+09:00 Masaki Murooka <notifications@github.com>:

mmurooka:
> as for cart position, you'd better to publish euslisp cart model with "Marker" 

Marker is also visualized. Please see carefully (especially model of caster is visible).
Video version (only visible for JSK member):
https://drive.google.com/a/jsk.imi.i.u-tokyo.ac.jp/file/d/0B5qanGnXorOuN2wyYml3M0I3Y2c/view?usp=sharing

> although that is not essential, putting AR marker is really bad ...

Thank you for many examples!
My goal is detecting object without any markers like this:
checkerboard is attached but not used.
vlcsnap-2016-10-07-11h38m01s70


I checked briefly and ir was not published in default settings. Maybe I'll need some setting to publish ir.

k-okada:
2016-10-07 11:40 GMT+09:00 Masaki Murooka <notifications@github.com>:

> My goal is detecting object without any markers like this:
>
Yes I know, and I think it is not difficult to create "general rectangle
tracker" that takes size of the rectangle, and that should be very useful
to track tables and doors, and useful to many cases,
So you can write another paper based on this method.

My concern is; when you move to a next step, working on tasks like
http://4.bp.blogspot.com/-cJxEnu4NpLA/T2L8SWkm9ZI/AAAAAAAAAT4/inB1i8texgQ/s1600/IMG_3303.JPG
,
http://blogs.c.yimg.jp/res/blog-f5-f0/fmyostaff/folder/545045/24/64590024/img_0
and
http://www.namegawa-shoga.jp/blog/wp-content/uploads/wpid-IMG_20140223_155227.jpg
.
and if we have only two choices, rectangle object and marker, are you going
to create anther recognition method? or back to marker?
If we have reliable, accurate method like marker, but do not draw so much
attention from viewers and that's helps a lot.


--
◉ Kei Okada

mmurooka:
Thank you for your advice. It becomes very helpful.
Texture based tracking is more general as you mention.

wkentaro:
I'm not sure what tf timestamp is.

PR #1458, opened 2016-10-07
Add interactive dynamic tf publisher

iory:
myimage


k-okada:
did you break backward compatibility?

--
◉ Kei Okada


2016-10-07 18:14 GMT+09:00 iory <notifications@github.com>:

iory:
it pointed out specific_transform_publisher.cpp and specific_transform_subscriber.cpp?
This nodes are not originally compiled.
But jsk_topic_tools depednd on dynamic_tf_publisher.
This is incorrect, so I removed it from package.xml on jsk_topicl_tools

k-okada:
it should be build when they first import this, so I'd like know why it is
disabled,
https://github.com/jsk-ros-pkg/jsk_common/commit/3860dbd5c1b3727ed38ad9231b1560de39029b0d#diff-04b2149ed070aec1d90adfc5151f74c9
and it seems we have python version of the dynamic_tf_publisher,
jsk_topic_tools version and your new version, If we write this feature on
our self, we need to stabilize the code
see my comment on this type of feature ->
https://sourceforge.net/p/jsk-ros-pkg/mailman/message/31121900/

--
◉ Kei Okada


2016-10-12 2:35 GMT+09:00 iory <notifications@github.com>:

iory:
I don't know why it is disabled.
My version's dynamic_tf_publisher has interactive_marker.

> we need to stabilize the code

what should I do?
Rewrite code in python?

k-okada:
please explain your design .
does your code just provide interactive marker and send any topic/service
to existing python node?
or did you write something similar to existing code ?

 > I don't know why it is disabled.

Please ask @YoheiKakiuchi


--
◉ Kei Okada


2016-10-13 21:26 GMT+09:00 iory <notifications@github.com>:

iory:
My program publish tf, provide interactive marker and change parameter by dynamic_reconfigure.
They are all-in-one.

k-okada:
if so, please consider extending current python code to provide interactive
marker features, or try to replace current node with same node name, same
API and so on...

--
◉ Kei Okada


2016-10-14 12:58 GMT+09:00 iory <notifications@github.com>:

wkentaro:
I don't have enough time for a detailed investigation, but I'd like to provide below comments

- this seems have similar feature as nodes in jsk_interactive_marker, so why not you place this there?
- [jsk_interactive_marker/marker_6dof](https://github.com/jsk-ros-pkg/jsk_visualization/blob/master/jsk_interactive_markers/jsk_interactive_marker/src/marker_6dof.cpp), [jsk_interactive_marker/transformable_object_server](https://github.com/jsk-ros-pkg/jsk_visualization/blob/master/jsk_interactive_markers/jsk_interactive_marker/src/transformable_server_sample.cpp) is same but this node is simpler, so if there is no nodes which has same feature in jsk_interactive_marker, I support you adding this to jsk-ros-pkg.
- Though I agree with adding this node, maybe you can firstly send PR to [interactive_markers](https://github.com/ros-visualization/interactive_markers) as a proposal.

IMO, the procedure you should do is

1. Ask to @YuOhara and @mmurooka if same featured node exists, seems know much about this kind of node with participation to DRC.
2. If not 1, check if same featured node exists in [jsk_interactive_marker](https://github.com/jsk-ros-pkg/jsk_visualization/tree/master/jsk_interactive_markers/jsk_interactive_marker).
3. Send proposal PR to [interactive_markers](https://github.com/ros-visualization/interactive_markers).
4. If not 2, send PR to [jsk_interactive_marker](https://github.com/jsk-ros-pkg/jsk_visualization/tree/master/jsk_interactive_markers/jsk_interactive_marker) to place this node there.

k-okada:
seems to moved to https://github.com/ros-visualization/interactive_markers/pull/32

Please remember
- adding new feature by adding more code
is ok, but what we have to consider is
- adding new feature by removing existing code
So please discuss with Interactive Marker users @YoheiKakiuchi, @YuOhara, @mmurooka

wkentaro:
I'd like to use this. @iory are there any code which is released or merged into master?

iory:
I think that you should use marker_6dof.
https://github.com/jsk-ros-pkg/jsk_visualization/blob/master/jsk_interactive_markers/jsk_interactive_marker/src/marker_6dof.cpp
If you do not want to visualize any shape, you can set object_type to anything other than "cube, sphere, line, mesh".

PR #1467, opened 2016-10-25
[jsk_data] Create softlink for extracted files in download_data

wkentaro:
- For multiple workspaces like in jenkins.

PR #1469, opened 2016-10-26
[jsk_data] Fix removing of symlink destination path

wkentaro:


k-okada:
I think this is follow up of #1467

wkentaro:
Yes. I think now it works.
Could you please release the next version of jsk_topic_tools?

PR #1475, opened 2016-11-04
[jsk_data] add TransformStamped in data_collection_server

knorth55:
support topic type TransformStamped


knorth55:
TransformStamped can be save in yaml with #1476

PR #1481, opened 2016-11-10
Resolve dependency on python-gdown with rosdep

h-kamada:


k-okada:
which script need this?

wkentaro:
https://github.com/jsk-ros-pkg/jsk_common/blob/master/jsk_data/src/jsk_data/download_data.py#L81

k-okada:
if you add new dependency, please check if that is released on fedra/debian -> https://github.com/ros/rosdistro/pull/13397

wkentaro:
I see.

BTW, is that impossible to stop releasing fedra/debian? (Do we need to support fedra/debian?)

k-okada:
> BTW, is that impossible to stop releasing fedra/debian? (Do we need to
support fedra/debian?)

no, that's defined in build.ros.org

--
◉ Kei Okada


2016-12-13 15:40 GMT+09:00 Kentaro Wada <notifications@github.com>:

PR #1482, opened 2016-11-14
[jsk_topic_tools] add tf_to_transform node

knorth55:
Usage:

```
$ rosrun jsk_topic_tools tf_to_transform.py -h
usage: tf_to_transform.py [-h] [--duration DURATION]
                          [parent_frame_id] [child_frame_id]

positional arguments:
  parent_frame_id       parent frame id
  child_frame_id        child frame id

optional arguments:
  -h, --help            show this help message and exit
  --duration DURATION, -d DURATION
                        Duration [s]: default=1.0
```

param:
`duration` : duration
`parent_frame_id`: parent frame id
`child_frame_id`: child frame id

Output:
`TransformStamped`

### DIFF
--- a/doc/jsk_topic_tools/scripts/tf_to_transform.rst
+++ b/doc/jsk_topic_tools/scripts/tf_to_transform.rst
@@ -1,0 +1,99 @@
+==================
+tf_to_transform.py
+==================
+
+
+What is this?
+=============
+
+Specific TransformStamped publisher.
+You specify parent_frame_id and child_frame_id, and this node publish `geometry_msgs::TransformStamped`.
+You can define `--duration` (default=0.1).
+
+Difference from tf_to_pose.py
+=============================
+
+TransformStamped msg carries parent_frame_id and child_frame_id.
+Otherwise, PoseStamped msg only carries parent_frame_id.
+
+Usage
+=====
+
+Command Line
+------------
+
+.. code-block:: bash
+
+    $ rosrun jsk_topic_tools tf_to_transform.py -h
+    usage: tf_to_transform.py [-h] [--duration DURATION]
+                              [parent_frame_id] [child_frame_id]
+
+    positional arguments:
+      parent_frame_id       parent frame id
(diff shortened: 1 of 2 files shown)


k-okada:
why you do not use tf?

```
$ rosmsg show geometry_msgs/Transform
geometry_msgs/Vector3 translation
  float64 x
  float64 y
  float64 z
geometry_msgs/Quaternion rotation
  float64 x
  float64 y
  float64 z
  float64 w

$ rosmsg show geometry_msgs/Pose
geometry_msgs/Point position
  float64 x
  float64 y
  float64 z
geometry_msgs/Quaternion orientation
  float64 x
  float64 y
  float64 z
  float64 w
```

knorth55:
i want to get Transform from one frame to the other frame, and in that case, we need to `lookupTransform` function.
usage is like this: https://github.com/knorth55/jsk_apc/blob/data-collection/jsk_2016_01_baxter_apc/launch/bimanual_stow_data_collection.launch#L107-L125

wkentaro:
Maybe `tf_to_pose.py` satisfies the needs.
https://github.com/jsk-ros-pkg/jsk_common/blob/master/jsk_topic_tools/scripts/tf_to_pose.py

knorth55:
I tried tf_to_pose, but posestamped msg have only one header, which is parent frame id.
But transform stamped  has both child and parent frame id.
That's why i use transformstamped

k-okada:
please add document about "when we use tf_to_pose and when we use
tf_to_transform"..

--
◉ Kei Okada


2016-11-17 13:55 GMT+09:00 Shingo Kitagawa <notifications@github.com>:

knorth55:
add document about difference from `tf_to_pose.py`
https://github.com/knorth55/jsk_common/blob/27be2285e523d21fe77035b14fcc245fa217d55a/doc/jsk_topic_tools/scripts/tf_to_transform.rst#difference-from-tf_to_posepy

PR #1491, opened 2016-11-30
Cope with xenial + arm for deb status table

wkentaro:

PR #1503, opened 2017-01-05
Cope with arm64 for deb status table

wkentaro:
For https://github.com/jsk-ros-pkg/jsk_common/pull/1491#issuecomment-270557750


k-okada:
Thank you

PR #1511, opened 2017-03-25
[jsk_data] add openni_record.launch openni_play.launch

furushchev:
Record only depth image instead of pointcloud, but publish pointcloud on play.
support raw / compressed image.

tested with astra s

example usage:

```bash
# record
roslaunch openni2_launch openni2.launch depth_registration:=true
roslaunch jsk_data openni_record.launch name:=icp_2d record_compressed_rgb:=true record_compressed_depth:=true rosbag_option:='--duration=120'

# play
roslaunch jsk_data openni_play.launch name:=`pwd`/icp_2d_2017-03-25-20-52-46.bag rosbag_option:='--clock --loop'
```

### DIFF
--- a/jsk_data/launch/openni_play.launch
+++ b/jsk_data/launch/openni_play.launch
@@ -1,0 +1,65 @@
+<launch>
+  <arg name="name" doc="full path to bag file" />
+
+  <arg name="camera_namespace" default="camera" />
+  <arg name="rosbag_option"    default="--clock" />
+  <arg name="compressed"       default="false" />
+  <arg name="use_xterm"        default="false" />
+  <arg name="use_gui"          default="false" />
+
+  <param name="use_sim_time" value="true" />
+
+  <arg name="remapping_namespace" default="$(arg camera_namespace)" unless="$(arg compressed)"/>
+  <arg name="remapping_namespace" default="camera_rosbag" if="$(arg compressed)" />
+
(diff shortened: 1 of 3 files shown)


k-okada:
why you do not use https://github.com/jsk-ros-pkg/jsk_common/blob/2.2.3/jsk_data/launch/common_record.launch

k-okada:
this is merged, but reverted. please re-create PR if you still needed

furushchev:
@k-okada I mean `common_record.launch` is a launch file for being included. And it does not have playing method, record compressed topics, or playing pointclouds from compressed topics.

k-okada:
> And it does not have playing method,

we have playing method for each robot, if you still need more, create
common_play.launch

https://github.com/jsk-ros-pkg/jsk_common/blob/2.2.3/jsk_data/launch/pr2_play.launch
https://github.com/jsk-ros-pkg/jsk_common/blob/2.2.3/jsk_data/launch/baxter_play.launch
https://github.com/jsk-ros-pkg/jsk_common/blob/2.2.3/jsk_data/launch/hrp2_play.launch

> record compressed topics, or playing pointclouds from compresse

I'm not sure why you need compressed topics, but these method should be
supported for all robots.

--
◉ Kei Okada


2017-03-28 22:13 GMT+09:00 Furushchev <notifications@github.com>:

PR #1523, opened 2017-05-28
[jsk_topic_tools][connection_based_nodelet] add isSubscribed method

furushchev:
To see current subscribing state of a nodelet.


furushchev:
@wkentaro I think it's better to define as inline method. Is it better to define in `.cpp`?

wkentaro:
> @wkentaro I think it's better to define as inline method. Is it better to define in .cpp?

Fine by me.

PR #1534, opened 2017-06-26
Support int vector parameters in rosparam utils

knorth55:


wkentaro:
This was for https://github.com/jsk-ros-pkg/jsk_recognition/pull/2131, and we no longer need that PR.
However, I think this PR is useful because user expects int vector also can be used for rosparam, but currently float and string are supported.

wkentaro:
I tested std::vector<int> can be gotten by `pnh_->param`.

k-okada:
I think we'd better to add vector of vector of integer too, btw 
> I tested std::vector can be gotten by pnh_->param.
means we do not need `jsk_topic_tools::read*` ? if so, can we warn "deprecated, use param...." ?

PR #1540, opened 2017-06-28
[jsk_data] support CameraInfoYAML in data_collection_server

knorth55:
Support camera_info YAML for data_collection_server.
Output format is as `camera_calibration`  `cameracalibrator.py` output.


wkentaro:
I'm not sure why you need this instead of saving it as `YAML`, could u please explain that?

knorth55:
This format can be used when launching camera with thi camera_info.

PR #1543, opened 2017-07-26
need to use system python, with travis-python

k-okada:
we can not find python module install by apt c.f. jsk-ros-pkg/jsk_3rdparty#117


k-okada:
move to #1548

PR #1544, opened 2017-08-09
jsk_topic_tools: add stealth_relay for silently subscribing topic

furushchev:
This node relays topic from `~input` and `~output` as normal `rostopic relay`, but it does not increment number of subscribers (actually increments number of subscribers, but not looks like so)


snozawa:
:+1:

furushchev:
rewrote in C++ to be overridden.

furushchev:
@k-okada Now changed following your review! :+1:

PR #1545, opened 2017-08-11
Fix undefined variable arg in rosview on bash

wkentaro:


k-okada:
`rosview` seems equal to `rosed`

```
EDITOR='emacs -nw' rosed euslisp common.l
EDITOR='less' rosed euslisp common.l
EDITOR='vi' rosed euslisp common.l
```

PR #1547, opened 2017-08-18
update .travis.yml to run on trusty environment

k-okada:


k-okada:
we alerady updated to 0.4.37 https://github.com/jsk-ros-pkg/jsk_common/pull/1579

PR #1558, opened 2017-09-05
jsk_topic_tools: connection_based_nodelet: fix typo in advertiseCamera

furushchev:
`advertiseCamera` with image_transport occurs infinite loop

### DIFF
--- a/jsk_topic_tools/include/jsk_topic_tools/connection_based_nodelet.h
+++ b/jsk_topic_tools/include/jsk_topic_tools/connection_based_nodelet.h
@@ -237,7 +237,7 @@ namespace jsk_topic_tools
                     int queue_size)
     {
       NODELET_WARN("advertiseCamera with ImageTransport is deprecated");
-      return advertiseCamera(nh, it, topic, queue_size);
+      return advertiseCamera(nh, topic, queue_size);
     }
     
     image_transport::CameraPublisher


furushchev:
`resized_image_transport<=0.3.21` hangs up nodelet manager due to this bug.

furushchev:
failure on hydro

```
executing command [apt-get install -y -qq python-requests]

WARNING: The following packages cannot be authenticated!

  python-requests

E: There are problems and -y was used without --force-yes

executing command [apt-get install -y -qq gawk]

WARNING: The following packages cannot be authenticated!

  libsigsegv2 gawk

E: There are problems and -y was used without --force-yes

executing command [apt-get install -y -qq bison]

WARNING: The following packages cannot be authenticated!

  libbison-dev bison

E: There are problems and -y was used without --force-yes
```

k-okada:
@furushchev please update requested  version number at https://github.com/jsk-ros-pkg/jsk_recognition/blob/master/resized_image_transport/package.xml#L33-L34

furushchev:
@k-okada I think we don't have to change `package.xml` if this fix is just released since `resized_image_transport>= 0.3.22` does not use this broken API anyway.

k-okada:
i see, thank you.

wkentaro:
@furushchev FYI, image_proc has `image_proc/resize` nodelet: http://wiki.ros.org/image_proc/

furushchev:
@wkentaro Wow, I didn't know that. Thx!

PR #1560, opened 2017-10-25
Support network interface name convention from ubuntu 15.10

furushchev:
Fixes https://github.com/jsk-ros-pkg/jsk_common/issues/1559
c.f.:
https://askubuntu.com/questions/702161/why-is-my-interface-now-wlp2s0-instead-of-wlan0


furushchev:
Oops. Wrong branch...

PR #1566, opened 2017-10-29
Fix travis push build: Use conda python

wkentaro:


wkentaro:
image

I have no idea, but there is no stdout outputs.

PR #1570, opened 2017-11-01
travis: drop hydro

furushchev:
According to offline talk with @k-okada, we're now considering about dropping hydro.
If this test does not pass, only drop hydro and keep `jsk_travis` unchanged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,4 +22,8 @@ env:
     - ROS_DISTRO=indigo USE_DEB=true BEFORE_SCRIPT='$CI_SOURCE_PATH/.travis_before_script_opencv3.bash'
     - ROS_DISTRO=jade USE_DEB=true
     - ROS_DISTRO=kinetic USE_DEB=true
+matrix:
+  fast_finish: true
+  allow_failures:
+    - env: ROS_DISTRO=hydro USE_DEB=false CATKIN_TOOLS_BUILD_OPTIONS='-iv --summarize --no-status --no-color'
 script: source .travis/travis.sh


furushchev:
test failed on hydro (it is dropped though)

```bash
+ rosdep install -q -y --rosdistro hydro -n -q -r --ignore-src --from-paths /opt/ros/hydro/share .
WARNING: Package "ompl" does not follow the version conventions. It should not contain leading zeros (unless the number is 0).
WARNING: Package "ompl" does not follow the version conventions. It should not contain leading zeros (unless the number is 0).
WARNING: Package "ompl" does not follow the version conventions. It should not contain leading zeros (unless the number is 0).
No packages found matching iproute2.
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
libcmt: No definition of [libopencv-dev] for OS version [precise]
jsk_data: No definition of [python-click] for OS version [precise]
jsk_tools: Cannot locate rosdep definition for [rosemacs]
No packages found matching iproute2.
E: Unable to locate package iproute2
No packages found matching iproute2.
ERROR: the following rosdeps failed to install
  apt: command [apt-get install -y -qq iproute2] failed
  apt: Failed to detect successful installation of [iproute2]
Continuing to install resolvable dependencies...
executing command [apt-get install -y -qq iproute2]
+ EXIT_STATUS=1
```

https://github.com/ros/rosdistro/pull/16232 will fix this issue but anyway I think it's ok to be merged. :+1:

wkentaro:
@furushchev Why did you decide to move hydro to `allow_failure` section, instead of dropping it completely?

PR #1576, opened 2018-01-14
Fix roscore regardless

furushchev:
- jsk_topic_tools: `isMasterAlive` was not working correctly.
- jsk_tools: `roscore_regardless.py`
  - Added feature to pass signals sent to parent to children
  - Set proper attributes on spawn child process
  - Clean and quiet log printing
  - Killing by gradually escalating signals like `roslaunch`

### DIFF
--- a/jsk_tools/bin/roscore_regardless.py
+++ b/jsk_tools/bin/roscore_regardless.py
@@ -1,56 +1,154 @@
 #!/usr/bin/env python
+# -*- coding: utf-8 -*-
 
-
-import rospy
-import time
-import subprocess
-import signal
 import os
+import rospy
+import signal
+import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


mmurooka:
@furushchev 
I might heard from you before that some update is necessary in the following use of roscore_regardless (maybe something about respawn?) after this PR is merged, but I forgot that. Do you remember that?
https://github.com/start-jsk/rtmros_hrp2/blob/master/jsk_hrp2_ros_bridge/scripts/hrp2_stat.sh#L13

furushchev:
@mmurooka Yes, previously a target node is respawned by default, which is now disabled.
If you want to enable respawning target node, I provide an option `--respawn` or `-r` for that.

PR #1579, opened 2018-05-03
jsk_topic_tools: add synchronized_throttle

furushchev:
Add `SynchronizedThrottle` nodelet which is like `topic_tools/throttle` but this nodelet accepts multiple topics as inputs and all throttled messages are synchronized.
Added a test code / documentation.

synchronized throttle nodelet


jsk_topic_tools/src/synchronized_throttle_nodelet.cpp
```diff
@@ -0,0 +1,418 @@
...
+      if (pub_[i].getNumSubscribers() > 0) {
+        subscribe();
+        subscribed_ = true;
+        break;
+      }
+    }
+  }
+}
+
+void SynchronizedThrottle::disconnectCb()
```
furushchev: Hmm, I just found that only if `approximate_sync` is enabled, unsubscription is delayed for 10~20 seconds and I have no idea about the behavior. (the function call `disconnectCb` itself is delayed)
But IMO, it is a minor issue and it is assumed that this nodelet is used mostly with exact synchronization, so still I'd like to be merged now.

furushchev:
Created pull request also in upstream: https://github.com/ros/ros_comm/pull/1391

furushchev:
@k-okada Here is one use case: https://github.com/jsk-ros-pkg/jsk_robot/pull/928

furushchev:
@k-okada Sorry I added wrong milestone, since 2.2.6 was still open after released.
I modified it to `2.2.7`.

PR #1585, opened 2018-06-20
jsk_topic_tools: add option to display diagnostic messages on warning level

furushchev:
This PR adds `~use_warning` parameter to display diagnostic messages on warning level instead of error level on `DiagnosticNodelet`.


furushchev:
The name `~use_warning` is too general, so any discussion is very welcome.

furushchev:
CC: @708yamaguchi

furushchev:
After discussion with active users, I changed to use `/diagnostic_nodelet/use_warn` for global ns, and `~use_warn` in local ns to set diagnostic status level on non OK status.

furushchev:
Added to milestone for possible next release

PR #1586, opened 2018-06-24
jsk_topic_tools/cmake/nodelet.cmake: add random prefix before _single…

k-okada:
…_nodelet_exec_name

I have asked personal question from a @sanketrahul about '`target with same name already exists`. This will occur only when you use `catkin_make`, `catkin build` users do not encounter this error, because it is isolated build so that current cmake target will not collide with another packages's target.
```
[0m[31m[1mCMake Error[0m at /opt/ros/kinetic/share/jsk_topic_tools/cmake/nodelet.cmake:26 (add_executable):
[0m  add_executable cannot create target "pointcloud_to_pcd" because another
[0m  target with the same name already exists.  The existing target is an
[0m  executable created in source directory
[0m  "/home/baxter2/pcl_ros_ws/src/perception_pcl/pcl_ros".  See documentation
[0m  for policy CMP0002 for more details.
[0m[36m[4mCall Stack (most recent call first):[0m
[0m  jsk_recognition/jsk_pcl_ros_utils/CMakeLists.txt:102 (jsk_nodelet)
```
https://gist.github.com/sanketrahul/1223fa186110499f4ef01533f928b6d9


iory:
@k-okada 
I checked ```build failed``` when we use cakin_build command in jsk_common.
Also, I checked ```build succeeded`` by applyging this patch.

k-okada:
@iory d70ebff is merged commit including a9f4f52d4a08f4d2c8805c142aa11adb2be2c79c 0397a78070182f1758c0c65524ff0be0cc4d94f1 0f99a574f1c22d65683458f1b46e685f79a1bff4 ...

iory:
I see. Thank you!

jsk_topic_tools/cmake/nodelet.cmake
```diff
@@ -4,7 +4,8 @@ macro(jsk_nodelet _nodelet_cpp _nodelet_class
     _single_nodelet_exec_name _CODE_PARAM_NAME)
   list(APPEND ${_CODE_PARAM_NAME} ${_nodelet_cpp})
   set(NODELET ${_nodelet_class})
-  set(DEFAULT_NODE_NAME ${_single_nodelet_exec_name})
+  string(RANDOM __NODENAME_PREFIX)
+  set(DEFAULT_NODE_NAME JSK_NODELET_${__NODENAME_PREFIX}_${_single_nodelet_exec_name})
```
furushchev: @k-okada Is there any rationale that you randomized the prefix?
I wonder any target misses caches in the case of `catkin build --force-cmake`.
How about changing `__NODENAME_PREFIX` to, for example, just `PROJECT_NAME`?

k-okada: @furushchev thank you for feedback, yes `PROJECT_NAME` would be better https://github.com/jsk-ros-pkg/jsk_common/pull/1591

PR #1591, opened 2018-08-02
use PROJECT_NAME instad of __NODENAME_PREFIX (RANDOM)

k-okada:
see https://github.com/jsk-ros-pkg/jsk_common/pull/1586/files#r207146300


furushchev:
@k-okada Thanks for fix!

PR #1594, opened 2018-08-08
update jsk_travis to 0.4.38 add lunar and melodic

k-okada:


k-okada:
travs passed, but we stil need https://github.com/ros/ivcon/issues/4, https://github.com/ros/convex_decomposition/issues/7, https://github.com/ros/rosdistro/pull/18767 for releasing.

PR #1608, opened 2018-11-02
check if the pkg exists and path is writable

k-okada:
Fix http://build.ros.org/view/Kbin_uX64/job/Kbin_uX64__jsk_pcl_ros_utils__ubuntu_xenial_amd64__binary/132/console


k-okada:
we need to release this immediately, because of http://build.ros.org/view/Ibin_uT64/job/Ibin_uT64__resized_image_transport__ubuntu_trusty_amd64__binary/126/console

PR #1621, opened 2019-04-09
[jsk_data] Add option for force overwrite decompressed bag

iory:
If we would like to decompress `hoge.bag`, `decompress_rosbag` function's result is `hoge.orig.bag` and `hoge.bag` which is decompressed rosbag.
However, if  `hoge.orig.bag`  already exists, the function not decompress the rosbag.
So this PR addes, `force` option to force decompress.

### DIFF
--- a/jsk_data/src/jsk_data/download_data.py
+++ b/jsk_data/src/jsk_data/download_data.py
@@ -60,11 +60,13 @@ def extract_file(path, to_directory='.', chmod=True):
     return root_files
 
 
-def decompress_rosbag(path, quiet=False, chmod=True):
+def decompress_rosbag(path, quiet=False, chmod=True, force=False):
     print('[%s] Decompressing the rosbag' % path)
     argv = [path]
     if quiet:
         argv.append('--quiet')
+    if force:
+        argv.append('--force')
     try:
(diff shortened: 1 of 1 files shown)


jsk_data/src/jsk_data/download_data.py
```diff
@@ -60,11 +60,13 @@ def extract_file(path, to_directory='.', chmod=True):
     return root_files
 
 
-def decompress_rosbag(path, quiet=False, chmod=True):
+def decompress_rosbag(path, quiet=False, chmod=True, force=True):
```
YutoUchimi: This is just a comment, but why is the default value of argument `force` different from https://github.com/jsk-ros-pkg/jsk_common/pull/1621/files#diff-47c43621e14db9c2464d84a6e6e6ee81R129 ?

iory: we should not change default behaviour, so i'll change it `force=False` in default.

k-okada:
@iory @YutoUchimi please reopen if we still need this feature, 
In addition to https://github.com/jsk-ros-pkg/jsk_common/pull/1626

PR #1625, opened 2019-04-21
Update package name so it does not produce extra lines in rospack list

J-Rojas:


J-Rojas:
I encountered this bug while running this demo:

https://github.com/tork-a/visualization_rwt/tree/kinetic-devel/rwt_plot.

roswww uses rospack list to inspect the package list in Python. However because of the extra new lines in this package name field, the roswww node dies. While their code should not crash, the package field should also conform to standards that the ros tools do not current normalize.

k-okada:
@J-Rojas thank you for the contribution!

PR #1633, opened 2019-07-12
Add debian in generate_deb_status_table.py

knorth55:
Current script only support for ubuntu, I update to support for debian.

jsk_common


knorth55:
@k-okada can you merge this?

PR #1636, opened 2019-09-14
[jsk_topic_tools] import _pickle as pickle for python3

knorth55:
cpickle is no more used in python3


knorth55:
I add a comment.

PR #1641, opened 2019-11-08
[jsk_data/common_record.launch] Add tf_static record

iory:
See #1639


itohdak:
LGTM.
I confirmed with hrp2017.

PR #1648, opened 2020-05-26
update jsk_travis 0.5.12, add noetic build test

k-okada:
see http://wiki.ros.org/noetic/Migration for migration info

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ env:
     - ROS_DISTRO=kinetic USE_DEB=true
     - ROS_DISTRO=lunar   USE_DEB=true
     - ROS_DISTRO=melodic USE_DEB=true
+    - ROS_DISTRO=noetic  USE_DEB=true
 script: source .travis/travis.sh
 after_success:
   # trigger build of jsk-docs.readthedocs.org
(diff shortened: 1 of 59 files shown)


knorth55:
after this modification, we cannot build on indigo because of package.xml format 3.
before that, i think it is better to have new release version for indigo robots.

knorth55:
> after this modification, we cannot build on indigo because of package.xml format 3.
before that, i think it is better to have new release version for indigo robots.

it is my misunderstanding, probably.
But, I remember that we could not build some packages with format 3 on Fetch.
@708yamaguchi do you remember it?

I need to figure out the problem for Fetch and PR2.

708yamaguchi:
@knorth55 

Do you mean https://github.com/fetchrobotics/fetch_ros/pull/120 ?

knorth55:
>after this modification, we cannot build on indigo because of package.xml format 3.
before that, i think it is better to have new release version for indigo robots.

I found out it is not indigo problem, but our fetch problem, so I move to jsk_robot.
https://github.com/jsk-ros-pkg/jsk_robot/issues/1221

>Do you mean fetchrobotics/fetch_ros#120 ?

@708yamaguchi Yes, that one. Thank you!

knorth55:
by the way, for python3, don't we need to use `catkin_install_python` as written in http://wiki.ros.org/UsingPython3/SourceCodeChanges?
In my understanding, we need to set it for the shebangs of python executable files.

PR #1691, opened 2021-06-11
updaet jsk_travis to 0.5.21

k-okada:
fix jsk_travis to 0.5.21 and add GithubActions indigo.yml  kinetic.yml  melodic.yml  noetic.yml  python3.yml

see https://github.com/k-okada/jsk_common/actions to if this works.


k-okada:
this should pass Travis, see #1688

PR #1694, opened 2021-06-17
Set ccache prefix only for cmake <= 3.4

furushchev:
Same fix as https://github.com/jsk-ros-pkg/jsk_visualization/pull/780
This PR is fix for error about ccache like reported below:
https://github.com/CoatiSoftware/Sourcetrail/issues/1081


furushchev:
I'm mainly using both bionic/melodic and focal/noetic.

furushchev:
Maybe need to restart testing on kinetic.
@k-okada 
I could not find a "Restart this test" button. Could you do it?

pazeshun:
@furushchev I restarted `.github/workflows/kinetic.yml / kinetic (pull_request)`.

PR #1696, opened 2021-07-01
[WIP][tf_relay] add tf_relay package

sktometometo:
Add tf_relay package.

This node broadcasts transforms from a given fixed frame to a given output frame which is the same as reference frame in specified duraiton even when reference frame is not broadcasted.

this is usefull if you want to use AR marker frame even when robot cannot see it.

## TODO List

- [ ] Add demo launch
- [ ] Add tests


sktometometo:
Moved from https://github.com/jsk-ros-pkg/jsk_common/pull/1698

PR #1697, opened 2021-08-20
[jsk_data] add wait_timer for timer method

shmpwk:
As to data_collection_server, wait_service_request for request method was implemented by https://github.com/jsk-ros-pkg/jsk_common/pull/1669 . 
So I apply it to the other methods such as timer method and all method.

### DIFF
--- a/jsk_data/node_scripts/data_collection_server.py
+++ b/jsk_data/node_scripts/data_collection_server.py
@@ -285,10 +285,10 @@ class DataCollectionServer(object):
 
     def wait_msgs_update(self):
         now = rospy.Time.now()
-        for msg_key in self.msgs.keys():
+        for msg_key in self.msg.keys():
             time_diff = None
             while time_diff is None or time_diff < 0:
-                stamp = self.msgs[msg_key]['stamp']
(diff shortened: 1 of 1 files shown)


knorth55:
I think `waiting` for `all` method is meaningless because all `msgs` are arrived and updated in `sync_sub_and_save_cb`.
but for `timer`, it might be useful.

knorth55:
@shmpwk can you check https://github.com/shmpwk/jsk_common/pull/1 ?

shmpwk:
@knorth55 You're right. I just tried it for `timer` and thought that it can be also applied to `all`.

shmpwk:
Thank you for the revise.
I fix typo on https://github.com/shmpwk/jsk_common/pull/1 by https://github.com/jsk-ros-pkg/jsk_common/pull/1697/commits/fac3f9c863f7298ff47868c3fee27010d8b7d03f. 
I confirm it works with `wait_save_request`: `true` or `false` and `wait_timer`: `true` or `false`.

knorth55 (approved):

PR #1705, opened 2021-11-09
[jsk_image_capture] add jsk_image_capture pacakge

sktometometo:
Add jsk_image_capture package.

# TODO

- [ ] Add README.md
- [ ] Add eusliep client example

### DIFF
--- a/jsk_image_capture/src/jsk_image_capture/jsk_image_capture.py
+++ b/jsk_image_capture/src/jsk_image_capture/jsk_image_capture.py
@@ -1,0 +1,69 @@
+# -*- encoding: utf-8 -*-
+
+import rospy
+from cv_bridge import CvBridge
+from sensor_msgs.msg import Image
+import cv2
+
+from .srv import Capture, CaptureRequest, CaptureResponse
+
+
+def capture_image(image_topic, file_name, cv_bridge):
+    duration_timeout = rospy.get_param('~duration_timeout', 10.0)
+    try:
+        msg = rospy.wait_for_message(
+                        image_topic,
+                        Image,
+                        timeout=rospy.Duration(duration_timeout)
+                        )
+        image = cv_bridge.imgmsg_to_cv2(msg, desired_encoding='bgr8')
+        cv2.imwrite(file_name, image)
+        return True, 'Success'
+    except rospy.ROSException as e:
+        rospy.loginfo('Error: {]'.format(e))
+        return False, 'Error: {}'.format(e)
+
+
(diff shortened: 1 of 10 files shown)


k-okada:
did you consider using https://github.com/ros-perception/image_pipeline/blob/noetic/image_view/src/nodes/image_saver.cpp ?

sktometometo:
@k-okada I know that node. But I need an interface to save an image with a specified file name. That node does not provide it.

k-okada:
you can set file name via param. It is not clean solution and you might not
like it. But if you create client software, it will hide raw interface.
https://github.com/ros-perception/image_pipeline/blob/71d537a50bf4e7769af513f4a3d3df0d188cfa05/image_view/src/nodes/image_saver.cpp#L207

--
◉ Kei Okada



2021年11月9日(火) 16:11 Koki Shinjo ***@***.***>:

knorth55:
i think the problem of `image_saver` is that it cannot dynamically change the filename.
the rosparameter is only loaded at the beginning.
I don't know if we should improve `image_saver` or make a new node.

k-okada:
If so, please consider adding that feature to https://github.com/jsk-ros-pkg/jsk_common/tree/master/jsk_ros_patch/image_view2
adding new feature to existing packages have many advantages;
- existing maintainer will also take care of your new feature
- your feature already have uses/system, so you have more chance to propagate your idea
Thus, even if you loose interest on that feature, someone will keep maintain it and will use them in their system. So you can put that results within your thesis and you can claim that your developed system is so useful/strong that people keep using them without original developers help.

jsk_image_capture/src/jsk_image_capture/jsk_image_capture.py
```diff
@@ -0,0 +1,69 @@
+# -*- encoding: utf-8 -*-
```
knorth55, line 1: change file name from `jsk_image_capture` to other (i.e. `jsk_image_capture_base`)
otherwise, python2 confuse importing `package` or `filename`.
this is the reason why you cannot import `jsk_image_capture.srv`

knorth55 (changes requested):

jsk_image_capture/src/jsk_image_capture/jsk_image_capture.py
knorth55: this is actually quite important tip, but not common.
do not put a file which has the same name as the package name.

sktometometo:
Moved scripts and service files to image_view2.

PR #1715, opened 2022-03-30
use rostopic.get_topic_list instead of master.getSystemState

98shimpei:
https://github.com/ros/ros_comm/commit/e96c407c64e1c17b0dd2bb85b67f388380527097#diff-21f05efee85a56a5b34825bb4fcfb53db99e38304206ddad7d50d6d4881696b2L1207
changes

```
-        state = master.getSystemState()
+        pubs, subs, _ = state
+        pubs, subs = get_topic_list(master=master)
```

Closes https://github.com/jsk-ros-pkg/jsk_common/issues/1714

### DIFF
--- a/jsk_tools/src/rostopic_host_sanity
+++ b/jsk_tools/src/rostopic_host_sanity
@@ -13,8 +13,7 @@ import rosgraph
 
 def main():
     master = rosgraph.Master('/ros_host_sanity')
-    state = master.getSystemState()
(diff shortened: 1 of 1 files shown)


knorth55 (changes requested):
can you check if `get_topic_list` exists or not and select which function can be called?

k-okada:
this is already merge as a part of #1717

PR #1735, opened 2022-05-26
[jsk_topic_tools/diagnostics_nodelet] Poke when start subscribing.

iory:
# What is this?

If the diagnostics nodelet is currently subscribed,but the publish function is not executed (`vital_checker->poke()` is not executed), it publishes diagnostics with an error level.
This publishes error level diagnostics only at the beginning because poke is not called when subscribing starts.

This is to change to output diagnostics at the OK level by inserting a poke when switching from the not subscribed state to the subscribed state.

### DIFF
--- a/jsk_topic_tools/include/jsk_topic_tools/diagnostic_nodelet.h
+++ b/jsk_topic_tools/include/jsk_topic_tools/diagnostic_nodelet.h
@@ -114,7 +114,11 @@ namespace jsk_topic_tools
      * True if summary is displayed as 'Warnings', otherwise it is displayed as 'Errors'
      */
     uint8_t diagnostic_error_level_;
-    
+
+    /** @brief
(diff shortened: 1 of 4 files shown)


708yamaguchi:
I did not debug yet, but `dual_fisheye_to_panorama` in `fetch1075` still publish diagnostics at the start of the subscribe.

iory:
Maybe https://github.com/jsk-ros-pkg/jsk_recognition/pull/2690 will fix the problem.

708yamaguchi:
This PR, https://github.com/jsk-ros-pkg/jsk_recognition/pull/2690 and https://github.com/jsk-ros-pkg/jsk_recognition/pull/2691 fix `dual_fisheye_to_panorama`'s error diagnostics at the start of the subscribe.

PR #1740, opened 2022-06-02
[jsk_topic_tools/connection_based_transport] Update the time of last_published_time

iory:
# What is this?

Update the time of `last_published_time` to make it possible to take the difference time between the time of start subscribing and the current time.

`ConnectionBasedTransport` checks if topic is input and output normally.
https://github.com/jsk-ros-pkg/jsk_common/blob/df29bfbe32b07045be2a8f2a951a091a4f3c7d1e/jsk_topic_tools/src/jsk_topic_tools/transport.py#L176-L195

`self._connection_status` changes to `SUBSCRIBED` when node are subscribed.
The `is_running` function checks `last_published_time` and is `True` if it is published within the `1 / vital_rate` time.

If it hasn't been subscribed for a while now, `is_running` will always be `False` at first because `last_published_time` will not be updated all the time when it is not subscribed.

Related to https://github.com/jsk-ros-pkg/jsk_common/pull/1735 (c++'s diagnostics problem)

Also, I added the docs for `diagnostics function`.
```
Users inherit this class and define publisher by calling the `self.advertise` function.
The return value of `self.advertise` is the `Publisher` class. User should execute publish in `callback` function.

### For Expert

When `publish` is executed, it is judged that this node is running normally, and published `diagnostics` becomes `OK` level.
However, if the publisher is subscribed but the `publish` function is not called, the node is not working properly and `diagnostics` will be at the `ERROR` level.

Some users may execute`publish` at any time (eg `rospy.Timer`).
In that case, by executing the `self.poke` function in the callback function, the state of `diagnostics` will be `OK` level without executing publish.
```

### DIFF
--- a/doc/jsk_topic_tools/class/python_connection_based_transport.md
+++ b/doc/jsk_topic_tools/class/python_connection_based_transport.md
@@ -5,6 +5,18 @@
 This class is a base-class which can start subscribing topics if published topics are subscribed by the other node.
 This is abstruct class.
 
+Users inherit this class and define publisher by calling the `self.advertise` function.
+The return value of `self.advertise` is the `Publisher` class. User should execute publish in `callback` function.
+
+### For Expert
+
+When `publish` is executed, it is judged that this node is running normally, and published `diagnostics` becomes `OK` level.
+However, if the publisher is subscribed but the `publish` function is not called, the node is not working properly and `diagnostics` will be at the `ERROR` level.
+
+Some users may execute`publish` at any time (eg `rospy.Timer`).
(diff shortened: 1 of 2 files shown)


708yamaguchi:
Thank you for your PR.

Currently, this feature is tested on fetch1075.
With this PR, fetch1075's `edgetpu_object_detector` does not publish diagnostics error at the beginning of subscribe.
https://github.com/jsk-ros-pkg/coral_usb_ros/blob/3ac4da97a7bd935469e7833a21a551275ae76c7f/python/coral_usb/detector_base.py#L68

708yamaguchi (approved):

708yamaguchi (approved):

PR #1743, opened 2022-06-07
[jsk_topic_tools/HzMeasureNodelet] Add diagnostics to monitor hz

iory:
# What is this?

Enable publishing diagnostics on `HzMeasureNodelet`.
This PR also adds a document.
The difference from `sanity_diagnostics.py` is that you can ignore the communication overhead when using it with nodelet.

```
This node publishes `hz` of target topic and `diagnostics`.

User can specify `~warning_hz` param.

If target `hz` is smaller than `~warning_hz`, this node outputs `diagnostics` at `WARN` level (if `~use_warn` is `True`) or `ERROR` level.
```

## Use case

For example, use it when you want to monitor the topic in the nodelet.

Screenshot 2022-06-07 23:11:17

At first, the topic (`/head_camera/depth_registered/hw_registered/image_rect_raw`) like this works with normal hz.
If hz drops for some reason, it will output error level diagnostics as follows:

Screenshot 2022-06-07 23:20:37

### DIFF
--- a/doc/jsk_topic_tools/class/hz_measure_nodelet.md
+++ b/doc/jsk_topic_tools/class/hz_measure_nodelet.md
@@ -1,0 +1,35 @@
+# HzMeasureNodelet (C++)
+
+## Description
+
+This node publishes `hz` of target topic and `diagnostics`.
+
+User can specify `~warning_hz` param.
+
+If target `hz` is smaller than `~warning_hz`, this node outputs `diagnostics` at `WARN` level (if `~use_warn` is `True`) or `ERROR` level.
+
+
+## Subscribing Topic
+
+- `~input` (`AnyMsg`):
+
+    Target topic.
+
+## Publishing Topic
+
+- `~output` (`std_msgs::Float32`):
+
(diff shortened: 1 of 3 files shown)


doc/jsk_topic_tools/class/hz_measure_nodelet.md
```diff
@@ -0,0 +1,35 @@
...
+
+- `~output` (`std_msgs::Float32`):
+
+    Target topic's `hz`.
+
+- `~diagnostics` (`diagnostic_msgs::DiagnosticArray`):
+
+    Diagnostic messages.
+
+## Parameter
```
708yamaguchi, line 28: You may forget to add `~warning_hz` description?

tkmtnt7000:
Nice feature!
I confirmed that this node works well in Fetch15 and Fetch1075.
```
[http://fetch15:11311][133.11.216.217] fetch@fetch15:~$ rostopic echo /head_camera/rgb/image_raw/measured_hz
data: 29.7914276123
---
data: 29.7830791473
---
```

PR #1745, opened 2022-06-22
[jsk_tools] Seperate sanity check frequency and diagnostics publish frequency

708yamaguchi:
### Problem

sanity diagnostics status in rqt_robot_monitor turns on and off.

### What I did

In this PR, I separate topic and node check frequency and diagnostics publish frequency.
- Check topic and node at low frequency to reduce CPU usage
- Publish diagnostics at high frequency to avoid the above problem.

Currently, when I run `roslaunch jsk_tools sample_sanity_diagnostics.launch`, `/diagnostics` topic is published at more than 1 Hz but `check_topic()` and `check_node()` methods are called at 0.1 Hz.

### DIFF
--- a/doc/jsk_tools/scripts/sanity_diagnostics.md
+++ b/doc/jsk_tools/scripts/sanity_diagnostics.md
@@ -50,9 +50,13 @@ This node publishes essential robot topic and node status to `/diagnostics`.
       - /roswww
     ```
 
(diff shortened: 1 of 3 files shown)


708yamaguchi:
@iory

Could you check this?

iory (approved):

PR #1750, opened 2022-07-05
[jsk_rosbag_tools] Fixed a bug when the specified fps is less than the fps of the topic in rosbag.

iory:
# What is this?

Now if the specified fps is lower than the hz of the topic in the rosbag, the video generation will not work.
https://drive.google.com/uc?id=1Py0x0Ys0O6M5O0D1oexYSrvEx-5z6Iue

This PR fixed a bug when the specified fps is less than the fps of the topic in rosbag.

- [x] Add test for fps.

### DIFF
--- a/jsk_rosbag_tools/python/jsk_rosbag_tools/bag_to_video.py
+++ b/jsk_rosbag_tools/python/jsk_rosbag_tools/bag_to_video.py
@@ -136,14 +136,14 @@ def bag_to_video(input_bagfile,
             if show_progress_bar:
                 progress.update(1)
             aligned_stamp = stamp - start_stamp
+            # write the current image until the next frame's timestamp.
             while current_time < aligned_stamp:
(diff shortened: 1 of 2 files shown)


708yamaguchi:
> the video generation will not work.

In today's fetch kitchen demo `Fetch kitchen patrol demo: 2022/07/05 (10:32:04)`, the actual demo time is 3 minutes.
However, the output panorama video time is 20 minutes.

This is because all recorded image rostopics are used to create the video.
Ideally, some rostopic should not be taken into the video according to the time stamp.
https://github.com/jsk-ros-pkg/jsk_common/blob/714713378fb037c9670bf48118010f1899c10bd5/jsk_rosbag_tools/python/jsk_rosbag_tools/bag_to_video.py#L145-L146

708yamaguchi (approved):

PR #1751, opened 2022-07-13
jsk_rosbag_tools: version of jsk_common is 2.2.12

k-okada:


iory:
I had mistakenly committed with jsk_rosbag_tools in 2.2.11.

I would be grateful if you could merge this PR after merging the following bug fix.
https://github.com/jsk-ros-pkg/jsk_common/pull/1750

PR #1753, opened 2022-07-14
[jsk_topic_tools] Add header file to check jsk_topic_tools' version

iory:
# What is this?

I add `version.h` for `jsk_topic_tools` to check version for build.
`jsk_topic_tools` has base class such as `DiagnosticNodelet` and `ConnectionBasedNodelet`.
For example, nodes in `jsk_recognition` are using those notelets.

This PR is useful to check `jsk_topic_tools` version.

Related to https://github.com/jsk-ros-pkg/jsk_recognition/pull/2712

PR #1755, opened 2022-07-15
[jsk_topic_tools/version] Generate jsk_topic_tools_version.h without calling catkin_package_xml.

iory:
# What is this?

Generate jsk_topic_tools_version.h without calling catkin_package_xml.
https://github.com/ros/catkin/blob/2612c4ce82a285cac13263576f8a85b8a6d6cd3d/cmake/catkin_package.cmake#L97-L100

Note that catkin_package_xml needs to be called, but it is called in catkin_package.

Related to https://github.com/jsk-ros-pkg/jsk_common/pull/1753


k-okada (approved):
Thanks!

PR #1756, opened 2022-07-28
[jsk_tools/sanity_diagnostics.py] Modified the timing of updating

iory:
# What is this?

Modified the timing of updating timestamp and states for each topic and each node.

Now all topics are managed in one timestamp, so when one topic is updated, the others are not checked.
This PR checks each topic correctly by maintaining a timestamp for each topic.

This PR is tested on PR1040 https://github.com/knorth55/jsk_common/pull/25


708yamaguchi (approved):
Thank you very much for your help.

I should have set `self.last_time_check_topic/node` variable for each topic/node.

PR #1757, opened 2022-08-23
[WIP] Add filtered relay

iory:
# What is this?

Add relay with filter function was added.


jsk_topic_tools/scripts/filtered_relay.py
```diff
@@ -0,0 +1,91 @@
...
+
+import roslib.message
+import rospy
+
+
+def expr_eval(expr, modules):
+    def eval_fn(topic, m, t):
+        return eval(expr, modules, {'m': m})
+    return eval_fn
+
```
nakane11, line 14: Would this function be divided as a module?
https://github.com/jsk-ros-pkg/jsk_common/search?q=expr_eval

jsk_topic_tools/scripts/filtered_relay.py
```diff
@@ -0,0 +1,91 @@
...
+                rospy.logerr("Expression using variables other than 'm': %s"
+                             % e.message)
+            except UnboundLocalError as e:
+                rospy.logerr('Wrong expression:%s' % e.message)
+            except Exception as e:
+                rospy.logerr(str(e))
+            else:
+                if not isinstance(res, (list, tuple)):
+                    res = [res]
+                self.pub_transform.publish(*res)
```
nakane11, line 66: I'm not sure which is better, but this node gets simpler without transform.

nakane11 (commented):
I checked this node works well.
I cannot come up with real example like https://github.com/jsk-ros-pkg/jsk_robot/pull/1570 right now because input and filter should be simple, but in such cases this node is useful.

jsk_topic_tools/scripts/filtered_relay.py
iory: That's a good direction.
I created a PR to divide it.
https://github.com/jsk-ros-pkg/jsk_common/pull/1760
Please check it.

jsk_topic_tools/scripts/filtered_relay.py
iory: I see.

> I'm not sure which is better

Let me know your opinion.

k-okada:
since https://github.com/jsk-ros-pkg/jsk_robot/pull/1570 has been closed and currently we do not have usage for this node. So may be we can leave this PR and re-start discussion / review until we found another use cases.

iory:
OK. We will wait until someone needs this node.

PR #1770, opened 2022-10-04
[ROS-O] do not specify c++ standard

v4hn:
it break with current log4cxx which requires c++17. No, going forward it's no option to specify the standard anymore. And it's not necessary either.

@k-okada

### DIFF
--- a/jsk_network_tools/setup.py
+++ b/jsk_network_tools/setup.py
@@ -1,6 +1,6 @@
 ## ! DO NOT MANUALLY INVOKE THIS setup.py, USE CATKIN INSTEAD
 
-from distutils.core import setup
+from setuptools import setup
 from catkin_pkg.python_setup import generate_distutils_setup
 
 # fetch values from package.xml
--- a/jsk_ros_patch/image_view2/CMakeLists.txt
+++ b/jsk_ros_patch/image_view2/CMakeLists.txt
@@ -11,12 +11,6 @@ if (CMAKE_VERSION VERSION_LESS 3.4)
   endif(CCACHE_FOUND)
 endif()
 
-if(NOT $ENV{ROS_DISTRO} STRLESS noetic)
-  add_compile_options(-std=c++14)
-elseif(NOT $ENV{ROS_DISTRO} STRLESS kinetic)
-  add_compile_options(-std=c++11)
-endif()
-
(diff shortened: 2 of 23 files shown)


v4hn:
I added multiple other patches needed to compile on debian testing.

k-okada:
@v4hn thanks for patches, i have createned new PR #1773 with CI test as well as backward compatibilies.  BTW, how did you setup ROS-O build environment, for example are you compile all ros tools from source? or are you using Debian-installed libraries?

here is my settings -> 
https://github.com/jsk-ros-pkg/jsk_common/pull/1773/files#diff-33cc4c925f7fd36575f5e5b61d1e9c942fea5189e2c67d09720d714e19151404

v4hn:
Great work!
Please watch my roscon talk linked from https://github.com/ros-o/ros-o if you didn't already.
There are three orthogonal ways to build with different pros/cons each.
CI testing against RoboStack and Debian can both make sense. Debian is more likely to fail as they include more upstream library changes.
Your approach is pretty much how to build on the Debian packages at the moment.
We still need a simple interface to overlay entries from rosinstall_generator with versions we patched in the ros-o organization because many patches are not merged upstream anymore (definitely not in a timely fashion)...
The core issue is that someone should write a tool to automatically extract a rosdistro file based on active ros-o repositories and hook it up with rosinstall_generator. That would help a lot. 
 Maintaining four different package origins (packaged, ros-gbp, upstream noetic-devel, ros-o) for a workspace is of course a burden, so I would personally go for an approach that does not use noetic upstream-develop at all, but only releases and ros-o branches of overlays exist and are up2date.
That also works better with the envisioned RoboStack integration of ros-o, where they don't consider upstream-devel at all.

v4hn:
That being said @jspricke worked on an approach recently to build debs directly from github actions and that might move more and more packages into debian and a custom deb repository in the future.

k-okada:
@v4hn thanks for explanation and I found that I missed your great talk at ROScon.
Integrating `rosinstall_generator` with ros-o and also deb build make sense to me too. Where should we discussion these isseus or find the current development status? https://github.com/ros-o/ros-o/discussions ??
I am also considering if we could contribute to provide computers to host deb files.

v4hn:
Please open a discussion there, yes. I can ping a few other involved people if there is something concrete to discuss there.

Issue #1789, opened 2023-08-31
[audio_video_recorder] sample_audio_video_recorder.launch doesn't work when audio_format is mp3 by default


Kanazawanaoaki:
When I tried to run the `audio_video_recorder` sample(`sample_audio_video_recorder.launch`), the video did not save properly.
The other sample `pr2_audio_video_recorder.launch` worked fine, so I'm trying to figure out the cause. As a conclusion, I think it does not work well if the `audio_format` is mp3.

Kanazawanaoaki:
I changed the default auido_format to wave format in https://github.com/jsk-ros-pkg/jsk_common/pull/1790.

For additional information, https://github.com/jsk-ros-pkg/jsk_recognition/tree/master/sound_classification#usage says that when dealing with audio device in ROS
By 
```
$ arecord -l
$ pactl list short sources
```
It is correct to investigate the values of arguments such as `device:=hw:1,0 channels:=1 depth:=16 sample_rate:=48000` and pass them as arguments to audio_capture.launch

Kanazawanaoaki:
I see.
Ideally, it would be better to read `audio_info` and specify parameters such as format from it.

From my tests, it seems that the `audio_video_recorder` does not save videos well for mp3 format audio topics, so if this is the case, I think it would be better to change the default so that the sample can be run with the default. If that is not the case, it may be OK to leave the mp3 as it is.

Sorry for the lack of debugging and understanding.

PR #1790, opened 2023-08-31
Change default audio_format to wave and Add device arg to sample audio_video_recorder

Kanazawanaoaki:
Related to https://github.com/jsk-ros-pkg/jsk_common/issues/1789
This PR changes the default `auido_format` of `sample_audio_video_recorder.launch` to wave format from mp3 so that it can be run by default, and adds `audio_device` argument so that a device can be specified.

### DIFF
--- a/audio_video_recorder/sample/sample_audio_video_recorder.launch
+++ b/audio_video_recorder/sample/sample_audio_video_recorder.launch
@@ -2,9 +2,10 @@
   <arg name="queue_size" default="100" />
   <arg name="file_name" default="/tmp/test.avi" />
   <arg name="file_format" default="avi" />
+  <arg name="audio_device" default="" />
   <arg name="audio_channels" default="1" />
   <arg name="audio_sample_rate" default="44100" />
-  <arg name="audio_format" default="mp3" />
+  <arg name="audio_format" default="wave" />
   <arg name="audio_sample_format" default="S16LE" />
   <arg name="video_height" default="480" />
   <arg name="video_width" default="640" />
@@ -20,6 +21,7 @@
   </node>
 
   <include file="$(find audio_capture)/launch/capture.launch">
+    <arg name="device" value="$(arg audio_device)" />
     <arg name="format" value="$(arg audio_format)" />
     <arg name="channels" value="$(arg audio_channels)" />
     <arg name="sample_rate" value="$(arg audio_sample_rate)" />


knorth55:
the format is defined in `audio_play` node.
`audio_play` node now publish the `audio_info` topic, and the topic format is written in the topic.
I haven't implemented it, but we can dynamically change the audio format and other params by `audio_info`.

knorth55:
the default format of `audio_common` 's `audio_capture` is `mp3`, so I think it is better to keep the default format as `mp3`.
but what do you think?
https://github.com/ros-drivers/audio_common/blob/836aa62522764ee7b8e4925b87ec7d84cfdd552c/audio_capture/src/audio_capture.cpp#L24

Kanazawanaoaki:
I see.
Ideally, it would be better to read `audio_info` and specify parameters such as format from it.

From my tests, it seems that the `audio_video_recorder` does not save videos well for mp3 format audio topics, so if this is the case, I think it would be better to change the default so that the sample can be run with the default. If that is not the case, it may be OK to leave the mp3 as it is.

Sorry for the lack of debugging and understanding.

knorth55:
I think I implmented to support mp3, too, but I'm not pretty sure.
can you give me the rosbag file which you tried and paste the `arecord` output for parameters?
I will try it when I have time.

but anyway, I think it is OK to change the default value to `wave`.
But in that case, can you add a note in `README.md` that default format of `audio_video_recorder` and `audio_capture` are different?

Kanazawanaoaki:
I understand that mp3 is also supported, then I felt that mp3 should be the default.  

This is the rosbag from sample_audio_video_recorder.launch in default mp3 format. 
https://drive.google.com/file/d/1--dAIJb97ch9ctBDqAP4Akb7bEWT0Img/view?usp=sharing
I tried it and it seems that `audio_play` can be saved to a file.

```
$ arecord -l
**** List of CAPTURE Hardware Devices ****
card 0: sofhdadsp [sof-hda-dsp], device 0: HDA Analog (*) []
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 0: sofhdadsp [sof-hda-dsp], device 1: HDA Digital (*) []
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 0: sofhdadsp [sof-hda-dsp], device 6: DMIC (*) []
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 0: sofhdadsp [sof-hda-dsp], device 7: DMIC16kHz (*) []
  Subdevices: 1/1
  Subdevice #0: subdevice #0
```

PR #1791, opened 2023-08-31
Add sample audio_recorder

Kanazawanaoaki:
If you want an audio file, [capture_to_file.launch  of audio_capture](https://github.com/ros-drivers/audio_common/blob/master/audio_capture/launch/capture_to_file.launch) is effective. However, since the audio topic is not published by capture_to_file.launch, a program that saves the audio topic to a file is required to simultaneously publish the topic and save the real-time audio file.

The people at https://stackoverflow.com/questions/63145510/how-to-write-ros-audiodata-message-into-wav-file are also looking for it, and there doesn't seem to be one, so I wrote a simple sample.  

If there is a demand and we need to do it properly, maybe we should understand gstreamer and split audio_video_recorder into audio only or video only (not many programs can make rostopic into a video with proper time?), or make it by bringing a part of `jsk_rosbag_tools` as a function.

### DIFF
--- a/audio_video_recorder/launch/audio_recorder.launch
+++ b/audio_video_recorder/launch/audio_recorder.launch
@@ -1,0 +1,19 @@
+<launch>
+  <arg name="audio_topic_name" />
+  <arg name="queue_size" />
+  <arg name="file_name" />
+  <arg name="channels" />
+  <arg name="depth" />
+  <arg name="sample_rate" />
+
+  <node name="audio_recorder" pkg="audio_video_recorder"
+        type="audio_recorder.py" output="screen">
+    <remap from="~input/audio" to="$(arg audio_topic_name)" />
+    <rosparam subst_value="true">
+      file_name: $(arg file_name)
+      channels: $(arg channels)
+      depth: $(arg depth)
+      sample_rate: $(arg sample_rate)
+    </rosparam>
+  </node>
+</launch>
--- a/audio_video_recorder/sample/sample_audio_recorder.launch
+++ b/audio_video_recorder/sample/sample_audio_recorder.launch
@@ -1,0 +1,25 @@
+<launch>
+  <arg name="file_name" default="/tmp/test_audio.wav" />
+  <arg name="channels" default="1" />
+  <arg name="depth" default="16" />
+  <arg name="sample_rate" default="16000" />
+  <arg name="sample_format" default="S16LE" />
+  <arg name="device" default="" />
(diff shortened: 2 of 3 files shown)


knorth55:
`audio_play` supports `filesink`, so we can directly save the audio data.

https://github.com/ros-drivers/audio_common/blob/836aa62522764ee7b8e4925b87ec7d84cfdd552c/audio_play/src/audio_play.cpp#L58-L79

Kanazawanaoaki:
Thank you very much! As it turns out, what I needed was to use `audio_play`'s `filesink`.

https://github.com/ros-drivers/audio_common/blob/836aa62522764ee7b8e4925b87ec7d84cfdd552c/audio_play/src/audio_play.cpp#L74-L79
```bash
roslaunch audio_play play.launch dst:=/tmp/output_audio_play.mp3 format:=wave
```

There are two situations where I think this feature is needed: 
one is when converting Robot's audio to a wav file instead of capturing the device and putting it into a speech recognition model that loads the wav file, 
```mermaid
sequenceDiagram
    participant R as Robot
    participant A as Audio Recorder
    participant M as Audio Recognition Model

    R->>A: audio ros topic
    A->>M: audio wave file
```
and the other is when saving the rostopic from audio_capture as a log in rosbag, but also saving it to a wav file and putting it into the model.
```mermaid
sequenceDiagram
    participant C as audio capture
    participant A as Audio Recorder
    participant M as Audio Recognition Model
    participant R as rosbag

    C->>A: audio ros topic
    A->>M: audio wave file
    C->>R: audio ros topic
```

knorth55:
You mean like these?
If so, it seems `audio_play` will solve the problem.
but the function is a bit difficult to find.
you can add the node directly to `audio_common`, if you want.

```mermaid
flowchart TD
  Robot --> |audio ros topic| audio_recorder
  audio_recorder -->|audio wave file| audio_recognition_model("Audio recognition model")
```

```mermaid
flowchart TD
  audio_capture --> |audio ros topic| audio_recorder
  audio_recorder -->|audio wave file| audio_recognition_model("Audio recognition model")
  audio_capture --> |audio ros topic| rosbag
```

Kanazawanaoaki:
That's right.
I don't know if this is necessary or the right solution, but I sent a PR to add record_to_file.launch to audio_common anyway.
https://github.com/ros-drivers/audio_common/pull/238

The problem is solved, so I'm closing this PR.

Issue #1798, opened 2024-04-04
somethign wrong on moviepy on 22.04


k-okada:
https://github.com/jsk-ros-pkg/jsk_common/actions/runs/8505002970/job/23324698471?pr=1797
```
[ 40%] Lock input requirements if they don't exist
cd /github/home/ws_current/src/jsk_common/jsk_rosbag_tools && /github/home/ws_current/build_isolated/jsk_rosbag_tools/catkin_generated/env_cached.sh rosrun catkin_virtualenv venv_lock /github/home/ws_current/build_isolated/jsk_rosbag_tools/venv --package-name jsk_rosbag_tools --input-requirements requirements.in --no-overwrite --extra-pip-args \"-qq\ --retries\ 10\ --timeout\ 30\"
    WARNING: Generating metadata for package pillow produced metadata for project name unknown. Fix your #egg=pillow fragments.
Traceback (most recent call last):
  File "/github/home/ws_current/build_isolated/jsk_rosbag_tools/venv/bin/pip-compile", line 8, in <module>
    sys.exit(cli())
  File "/usr/lib/python3/dist-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3/dist-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3/dist-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/github/home/ws_current/build_isolated/jsk_rosbag_tools/venv/lib/python3.10/site-packages/piptools/scripts/compile.py", line 392, in cli
    results = resolver.resolve(max_rounds=max_rounds)
  File "/github/home/ws_current/build_isolated/jsk_rosbag_tools/venv/lib/python3.10/site-packages/piptools/resolver.py", line 169, in resolve
    has_changed, best_matches = self._resolve_one_round()
  File "/github/home/ws_current/build_isolated/jsk_rosbag_tools/venv/lib/python3.10/site-packages/piptools/resolver.py", line 274, in _resolve_one_round
    their_constraints.extend(self._iter_dependencies(best_match))
[...]

PR #1799, opened 2024-04-04
[jsk_rosbag_tools] Fixed pillow version

iory:
Related to #1798 

In the course of addressing a build failure related to dependency resolution within our `jsk_rosbag_tools` package, we encountered an error indicating that `pip-compile` was unable to properly identify the version of the `Pillow` package required by `imageio==2.34.0` as part of the `moviepy==1.0.3` dependency. The error specifically pointed out that the package name or version was not recognized properly, which led to a halt in the dependency resolution process and subsequently the build process.

To resolve this issue, we've explicitly pinned the `Pillow` version to `8.4.0` in the `requirements.in` file. This action was taken after thoroughly investigating the underlying problem, which suggested that an explicit declaration could help `pip-compile` in resolving the correct package version and thus, facilitate a smoother dependency resolution process. Here's the commit that addresses this problem:

### DIFF
--- a/jsk_rosbag_tools/requirements.in
+++ b/jsk_rosbag_tools/requirements.in
@@ -1,3 +1,4 @@
+Pillow==8.4.0
 moviepy==1.0.3
 scipy<1.9.0
 numpy>=1.21.0


k-okada (approved):
CI passed!!!

image

PR #1814, opened 2025-02-21
enable noble techfak ros-o test

mqcmd196:


mqcmd196:
>AttributeError: module 'pkgutil' has no attribute 'ImpImporter'. Did you mean: 'zipimporter'?

would be fixed with 
- https://github.com/locusrobotics/catkin_virtualenv/pull/115

mqcmd196:
I'm not sure this is the correct solution but I added the temporary commit to fix ros-o/Noble CI.

I added @furushchev 's patch for fixing catkin_virtualenv and testing this on ros-o/Noble. After the newer catkin_virtualenv is released, we can remove the step to clone the @furushchev 's one.

PR #1607, opened 2018-10-29
[jsk_tools] Add audible warnings node

iory:
# What is this?
This is a general node that subscribes to `/diagnostics_agg` to speak the error content.
Robots using diagnostics can use this node.

## Target Action

* `/robotsound` (`sound_play/SoundRequestAction`)

    Target action name.

    If it is a different server name, please remap it like `<remap from="/robotsound" to="/sound_play" />`.

## Subscribing Topics

* `/diagnostics_agg` (`diagnostic_msgs/DiagnosticArray`)

    Aggregated diagnostics.

## Parameter

* `~speak_rate` (`Float`, default: `1.0`)

    Rate of speak loop.

* `~speak_interval` (`Float`, default: `120.0`)

    The same error will not be spoken until this number of seconds has passed.

* `~volume` (`Float`, default: `1.0`)

    Volume of speaking.

* `~language` (`String`, default: `""`)

    Language parameters for `arg2` in `sound_play/SoundRequestAction`.

* `~wait_speak` (`Bool`, default: `True`)

    If `True`, wait until finish saying one error.

* `~seconds_to_start_speaking` (`Float`, default: `0.0`)

    It is the time to wait for the node to speak after it has started.

    This is useful for ignoring errors that occur when the robot starts.

* `~speak_warn` (`Bool`, default: `True`)

    If `True`, speak warning level diagnostics.

* `~speak_error` (`Bool`, default: `True`)

    If `True`, speak error level diagnostics.

* `~speak_stale` (`Bool`, default: `True`)

    If `True`, speak stale level diagnostics.

* `~speak_when_runstopped` (`Bool`, default: `True`)

    If `True`, speak an error even if runstop is `True`.

* `~run_stop_topic` (`String`, default: `None`)

    Subscribe this topic if this value is specified.

* `~run_stop_condition` (`String`, default: `m.data == True`)

    Returning bool value condition using the given Python expression.
    The Python expression can access any of the Python builtins plus:
    ``topic`` (the topic of the message), ``m`` (the message) and ``t`` (time of message).

    For example, ``~run_stop_topic`` is ``robot_state (fetch_driver_msgs/RobotState)`` and if you want to check whether a runstop is a pressed, you can do the following.

    ```bash
    run_stop_condition: "m.runstopped is True"
    ```

- `~blacklist` (`Yaml`, required)

    User must always specify `name`. You can specify `message` as an option.

    These values are matched using python regular expressions.

    It is something like below:

    ```
    <!-- speak warning -->
[...]

PR #1821, opened 2026-05-20
[CI] fix ModuleNotFoundError: No module named 'pkg_resources' in Ubuntu 24.04

mqcmd196:


mqcmd196:
for

- https://github.com/jsk-ros-pkg/jsk_common/pull/1607

mqcmd196:
@k-okada As ros-o is released from techfak, we don't need to test jsk_common using full upstream sources. I think it can be removed from the workflows. What do you think?

mqcmd196:
@k-okada CI green

PR #1820, opened 2026-02-05
[ros2][jsk_topic_tools] Add ConnectionBasedLifecycleNode (ConnectionBasedNodelet in ROS1)

mqcmd196:
- **touching AMENT_IGNORE some packages**
- **[ros2][jsk_topic_tools] impl connection_based_lifecycle**
- **[jsk_topic_tools] allow both rclcpp_lifecycle, rclcpp publisher in connection_based_lifecycle_node**
- **[jsk_topic_tools] relax clang-format test**

PR #1822, opened 2026-05-28
Ros2 connection based

k-okada:
moved from https://github.com/jsk-ros-pkg/jsk_common/pull/1820

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -4,14 +4,9 @@ Standard: c++17
 
 # -- Indentation --
 IndentWidth: 2
-TabWidth: 2
 UseTab: Never
-ContinuationIndentWidth: 2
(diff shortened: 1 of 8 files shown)


mqcmd196:
@k-okada 
Thank you. If you separate the branch, let me clean the if/endif code in CMake / package.xml

k-okada:
Oh
If we can share same branch, that’s would be better


--
◉ Kei Okada


2026年5月28日(木) 11:12 Yoshiki Obinata ***@***.***>:

k-okada:
Let’s discuss later


--
◉ Kei Okada


2026年5月28日(木) 12:04 Kei Okada ***@***.***>:

PR #1826, opened 2026-09-11
Add ROS2 support to jsk_ros1_ros2_compat, dynamic_tf_publisher, jsk_topic_tools, jsk_tools, audio_video_recorder, jsk_data, image_view2, jsk_rosbag_tools

k-okada:
## Summary

Adds ROS2 support to several jsk_common packages, on top of the existing pr-1820 ROS2 CI groundwork:

- **jsk_ros1_ros2_compat** (new package): shared ROS1(roscpp/rospy)/ROS2(rclcpp/rclpy) compatibility helpers (message/service type aliasing, time/duration, publisher/subscriber/service-server templates, logging macros, tf2 buffer/listener/broadcaster construction, rosbag I/O), so individual packages don't each maintain their own copy.
- **dynamic_tf_publisher**: tf_publish.py/tf_set.py/tf_reconfigure_publisher.py ported to tf2_ros; ROS2 support added.
- **jsk_topic_tools**: transform_merger/static_transform_pose_stamped ported from tf to tf2_ros/geometry_msgs; log_utils.py's jsk_log*() functions consolidated into jsk_ros1_ros2_compat; ROS2 support added.
- **jsk_tools**, **audio_video_recorder**, **jsk_data**, **image_view2**, **jsk_rosbag_tools**: ROS2 support added, using jsk_ros1_ros2_compat throughout.

Each package keeps a single package.xml/CMakeLists.txt branched on `$ROS_VERSION`, with the ROS1-original lines left untouched (byte-identical, verified via diff) to minimize the diff against master.

Where a ROS2 port required a genuinely separable change, it's split into its own commit ahead of the actual "add ROS2 support" commit:
- tf1 → tf2_ros/geometry_msgs migrations (no functional change under ROS1)
- `Header` → `std_msgs/Header` field qualification in a few .msg/.srv files (required by ROS2's rosidl_adapter; verified via `genmsg.gentools.compute_md5()` that this does not change the ROS1 message/service MD5 sum, so it's not a wire-compatibility break)
- a templatization-only refactor in image_view2 (no functional change under ROS1)

image_view2's `ImageMarker2.msg` needed two further ROS2-only changes that *would* break the ROS1 MD5 (uppercase letters in field names are invalid under ROS2's rosidl, and rosidl has no `duration` primitive) -- rather than editing the shared ROS1 `msg/ImageMarker2.msg`, a ROS2-only variant lives at `ros2/msg/ImageMarker2.msg`, with the original ROS1 line commented out above each changed field.

## Test plan

- [x] `colcon build` (ROS2 Jazzy) of all 9 touched packages, clean from scratch, with `--symlink-install`
- [x] `g++ -fsyntax-only` against real ROS1 (ros-one distro) message/service headers (via gencpp) for all touched C++ files (image_view2, audio_video_recorder, jsk_topic_tools) -- caught and fixed two ROS1 compile-breaking bugs (`ImageMarker2`'s `points_3d`/`position_3d`/`lifetime_sec`/`lifetime_nsec` fields being referenced unconditionally instead of per-`$ROS_VERSION`)
- [x] `python3 -m py_compile` / real ROS2 import checks for all touched Python files
- [x] Verified via `genmsg.gentools.compute_md5()` that every ROS1 `.msg`/`.srv` field-qualification change in this PR leaves the ROS1 message/service MD5 sum unchanged


mqcmd196 (approved):
Let's merge and wait for feedback. I think there is no breaking in pre-exist features.

